PR #12, opened 2016-07-06
Added Viewing of Gcode Feature

BarbourSmith:
Added the ability to view the text of the loaded gcode file and created a scrollable text label which may be of use elseware also


UIElements/scrollableLabel.py
```diff
@@ -0,0 +1,6 @@
+from kivy.uix.label         import Label
+from kivy.uix.scrollview    import ScrollView
+from kivy.properties        import StringProperty
+
+class ScrollableLabel(ScrollView):
```
BarbourSmith: Needs a block comment

cncgc.py
```diff
@@ -97,6 +96,7 @@ def build(self):
         Initializations
         '''
         self.frontpage.gcodecanvas.initialzie()
+        self.showFeatures(6)
```
BarbourSmith: this is a bug fix, remove

PR #18, opened 2016-07-11
Add settings

BarbourSmith:


UIElements/screenControls.py
```diff
@@ -3,6 +3,7 @@
 from UIElements.otherFeatures                  import   OtherFeatures
 from UIElements.softwareSettings               import   SoftwareSettings
 from DataStructures.makesmithInitFuncs         import   MakesmithInitFuncs
+from kivy.config                               import ConfigParser
```
BarbourSmith: remove, this isn't needed here

UIElements/screenControls.py
```diff
@@ -30,7 +31,10 @@ def show_settings(self):
         Creates a new pop-up which lets the user see and edit the program settings
         
         '''
+        
+
```
BarbourSmith: this function never runs now and should be removed

UIElements/softwareSettings.py
```diff
@@ -1,4 +1,5 @@
-from kivy.uix.screenmanager import Screen
+from kivy.uix.settings                import SettingsWithSpinner
```
BarbourSmith: not needed

cncgc.py
```diff
@@ -37,6 +37,25 @@
 
 class GroundControlApp(App):
     
+    json = '''
```
BarbourSmith: it would be really cool if this weren't in the main file but its ok for now

settingstest.py
```diff
@@ -0,0 +1,117 @@
+"""
```
BarbourSmith: remove this file

PR #23, opened 2016-07-15
Textmessagescleanup

BarbourSmith:


BarbourSmith:
After mush strife for little change, front page text is improved

PR #25, opened 2016-07-27
Units displayed in real values

BarbourSmith:
removed some confusing things about how units are displayed

### DIFF
--- a/cncgc.py
+++ b/cncgc.py
@@ -80,7 +80,7 @@ class GroundControlApp(App):
         Scheduling
         '''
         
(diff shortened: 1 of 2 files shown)


cncgc.py
```diff
@@ -80,7 +80,7 @@ def build(self):
         Scheduling
         '''
         
-        Clock.schedule_interval(self.runPeriodically, .01)
+        Clock.schedule_interval(self.runPeriodically, .001)
```
BarbourSmith: change this back to .01

PR #32, opened 2016-08-09
General fixes

BarbourSmith:


cncgc.py
```diff
@@ -10,6 +10,7 @@
 from kivy.core.window           import Window
 from kivy.uix.button            import Button
 from kivy.clock                 import Clock
+from kivy.config                import Config
```
BarbourSmith: this is not needed

cncgc.py
```diff
@@ -74,6 +77,7 @@ def build(self):
         
         self.nonVisibleWidgets = NonVisibleWidgets()
         
+
```
BarbourSmith: get rid of extra space

PR #54, opened 2017-01-18
fix dots moving

BarbourSmith:
The indicators for z-axis motions didn't move with the rest of the gcode when the zero was redefined...now they do


BarbourSmith:
Test with opening a file

PR #68, opened 2017-01-31
Avoid crash at end of gcode file

blurfl:
Using step controls should avoid stepping past the end of the gcode
file.


blurfl:
Looks like this PR has the previous one wrapped in with it, sorry.

blurfl:
Sorry, again...

BarbourSmith:
Looks great to me. :-) 👍

PR #74, opened 2017-02-09
Use connection flag to toggle reconnect

BarbourSmith:
Use connnection status flag to determine if auto-reconnect runs

### DIFF
--- a/Connection/serialPort.py
+++ b/Connection/serialPort.py
@@ -25,7 +25,7 @@ class SerialPort(MakesmithInitFuncs):
         Runs on creation, schedules the software to attempt to connect to the machine
         
         '''
-        Clock.schedule_interval(self.openConnection, 2)
+        Clock.schedule_interval(self.openConnection, 5)
     
     def setPort(self, port):
         '''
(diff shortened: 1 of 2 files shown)


BarbourSmith:
Not ready to merge, issues are

-On Mac if the USB is pulled out after connecting it will reconnect automatically, but it disconnects and then reconnects a few times which is weird
-On Mac it will remember and automatically reconnect to the SAME usb port, but connecting to a new USB port requires selecting a different port
-On Mac turning on the drill which has a power cord running parallel to the USB cord causes the connection to drop. This probably has more to do with the 15foot USB extender plugged into a 4 foot USB cord than with the software but I want to explore it further. The same issue isn't seen on windows laptop?
-Much cleanup is needed

Connection/serialPort.py
```diff
@@ -6,6 +6,7 @@
 import sys
 import serial
 import threading
+import time
```
BarbourSmith: delete

Connection/serialPort.py
```diff
@@ -75,12 +76,17 @@ def openConnection(self, *args):
...
         #It only needs to be run once, it is run by connecting to the machine
         
-        #self.data.message_queue is the queue which handles passing CAN messages between threads
-        x = SerialPortThread()
-        x.setUpData(self.data)
-        self.th=threading.Thread(target=x.getmessage)
-        self.th.daemon = True
-        self.th.start()
+        print "connection Status"
+        print self.data.connectionStatus
```
BarbourSmith: delete

Connection/serialPortThread.py
```diff
@@ -75,6 +75,10 @@ def getmessage (self):
                 if len(msg) > 0:
                     
                     self.lastMessageTime = time.time()
+                    print "message length: "
+                    print len(msg)
```
BarbourSmith: cleanup

Connection/serialPortThread.py
```diff
@@ -99,6 +103,8 @@ def getmessage (self):
                             print "Gcode Ended"
                 
                 #check for serial connection loss
+                print "Downtime: "
```
BarbourSmith: cleanup

BarbourSmith:
@blurfl I ended up going with a flag because it turned out there was already a connection status flag. I'm not quite ready to merge the pull request because the behavior isn't quite what I want it to be, but it seems much improved from what it was, at least on my computer. I am testing on OS X 10.11.6.

If you have a second to check the behavior on your computer it would be much appreciated.

Please ignore the huge number of commits, I was editing on one computer and testing on the other so every time I changed anything it was a commit.

PR #77, opened 2017-02-23
Typo in gcodeCanvas.py

johnbenweber:
Fixed a typo which caused the toolpath to not be displayed in groundcontrol when the G21 command was encountered while parsing a gcode file.


BarbourSmith:
👍 👍 👍 Thanks for catching that!

PR #80, opened 2017-02-24
Workaround for line limitation in ScrollableTextPopup

johnbenweber:
I found that long Gcode files weren't showing up in the Actions>View Gcode modal window. It scrolls but is empty (first image). Through a process of elimination I found the maximum is ~450 lines though it appears to vary based on line length (so it's really probably a size issue rather than a number of lines issue). I'm not familiar enough with kivy to determine exactly why this is the case so I'm proposing a workaround where the Gcode is just truncated to 450 lines with some ellipses at the end (second image).

empty

truncated

### DIFF
--- a/UIElements/viewMenu.py
+++ b/UIElements/viewMenu.py
@@ -96,8 +96,12 @@ class ViewMenu(GridLayout, MakesmithInitFuncs):
         if len(self.data.gcode) is 0:
             popupText =  "No gcode to display"
         else:
-            for gcodeLine in self.data.gcode:
-                popupText = popupText + gcodeLine + "\n"
+            for lineNum, gcodeLine in enumerate(self.data.gcode):
+                if lineNum<447:
+                    popupText = popupText + gcodeLine + "\n"
+                else:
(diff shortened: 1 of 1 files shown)


blurfl:
I've hit this limit too. It looks like the whole file gets read into popupText. Searching online I see some warning about exceeding the 'gnu texture size limits' producing a black display. Your fix addresses that simply.

 I didn't find any example of scrolling through a text file using Kivy. I wonder if there is a need to see more than the beginning of the file in groundcontrol, maybe even fewer than 440 lines? One could go look through the file using some other program than ground control...

BarbourSmith:
👍 👍 👍  I only ever used this feature as a quick way to look at a gcode file to see if there was something really funky going on (no spaces or something like that). If you need to look at more than 450 lines, a regular text editor is a better bet.

Great pull request.

johnbenweber:
@blurfl that's interesting, I couldn't find anything relevant. I agree with you both that it's not imperative to be able to preview the entirety of a really long file. It's actually pretty inconvenient because scrolling appears to also zoom the gcode canvas.

PR #82, opened 2017-02-24
Master

blurfl:
When running for the first time, or without a gcode file selected, or
when the gcode file cannot be found, the 'openFile’ dialog starts
within the file structure of groundcontrol itself. This can be
confusing.
Instead, use the user’s home directory as the starting point. A
two-line change to viewMenu would set the path to path.expanduser('~')
if the gcode file is not found.


BarbourSmith:
Looks great 💯

PR #83, opened 2017-02-24
Fix define origin moving gcode multiple times

BarbourSmith:


BarbourSmith:
Repositioning a file on the sheet multiple times would cause an accumulated error. This was caused by an earlier fix which reset the gcode offset when a new file was loaded. The solution was to trigger the offset to reset when the file path changed, not when the gcode lines were updated

UIElements/frontPage.py
```diff
@@ -283,8 +283,7 @@ def moveOrigin(self):
         
         if self.data.units == "INCHES":
             amtToShiftX = self.numericalPosX - self.shiftX
-            amtToShiftY = self.numericalPosY - self.shiftY
-            self.shiftX = self.shiftX + amtToShiftX
+            amtToShiftY = self.numericalPosY - self.shiftYself.shiftX = self.shiftX + amtToShiftX
```
BarbourSmith: fix this

PR #85, opened 2017-02-28
Found a typo, doesn't appear to affect anything...

johnbenweber:
...except my OCD tendencies. I did a recursive grep search of the GroundControl directory just to make sure that these are the only references to the function in question.


BarbourSmith:
👍 👍 👍  Good catch!

PR #86, opened 2017-02-28
Open gcode list view

johnbenweber:
I added buttons to the open file popup to enable switching between icon and list view. The reason being that I was having trouble differentiating between files with long names in the icon view. I assume the icon view was initially selected intentionally thus the buttons to preserve that while also allowing the list view. In the kivy documentation ([link](https://kivy.org/docs/api-kivy.uix.filechooser.html?highlight=filechooserlistview#kivy.uix.filechooser.FileChooserListView)) the file names in list view are left justified but for some reason mine show up centered. I think left justified looks much better but couldn't figure out how to easily change it.

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -38,7 +38,7 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
     yPosition = 0
     zPosition = 0
     
-    def initialzie(self):
+    def initialize(self):
         with self.scatterObject.canvas:
             Color(1, 1, 1)
             
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -221,7 +221,8 @@
             on_press: root.show_gcode()
         Button:
             text: 'Reset Origin'
-            disabled: True
+            disabled: False
+            on_press: root.show_gcode()
                 
 <RunMenu>:
     GridLayout:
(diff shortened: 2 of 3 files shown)


BarbourSmith:
Looks great. 

I switched to the icon view because I thought it would be more intuitive, but you are right that the truncated file names can be hard to tell apart, especially when working with multiple version of the same file.

blurfl:
Yes! List view is very useful for longer file names. If you're good with Kivy, what do you think about using FileBrowser instead of LoadDialog? I'm poking at it, but not making headway. It would present a more polished interface, I think.

johnbenweber:
Hacking on the low hanging fruit of GroundControl is the first I've used kivy. Based on some searching it looks like you need to install Kivy Garden using pip ([link](https://kivy.org/docs/api-kivy.garden.html)). Garden appears to be a package manager for Kivy addons. Once you have that you'll have to install the FileBrowser addon ([link](https://kivy-garden.github.io/garden.filebrowser/api.html)). I'm sure you've seen both of those links but what I haven't seen is an example of how to implement the FileBrowser in kv language. Some other links I've found ([link1](http://stackoverflow.com/questions/32439330/add-filebrowser-in-kivy)) ([link2](https://github.com/kivy-garden/garden.filebrowser)).

While the FileBrowser looks great, my main concern would be the added dependencies/complexity but perhaps once GroundControl gets packaged up this becomes a non-issue. You could alternatively take the source from link2 above and just incorporate that into GroundControl since at its core FileBrowser just appears to be FileChooser with some formatting and helper functions.

PR #90, opened 2017-03-03
remove indicator ring

BarbourSmith:


BarbourSmith:
Only removes target indicator, position indicator is still there. Confusing name for PR

PR #92, opened 2017-03-07
Removed reference to target position and prevented stepping outside file bounds.

johnbenweber:
I found that since the target position indicator got removed, stepping through the gcode file with the arrow buttons caused GroundControl to crash. I removed the reference to targetPosition and all is well. 

While I was messing around I noticed you could get line negative line numbers and completion percentages. You could also get completion percentages above 100 and line numbers greater than the number of lines in the file. So I added a case structure to enforce the boundaries of the gcode file. Since we can't get outside the gcode file bounds I think the try...except structure is unnecessary so I removed it.

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -114,41 +114,18 @@ class FrontPage(Screen, MakesmithInitFuncs):
         self.shiftY = 0
     
     def moveGcodeIndex(self, dist):
-        self.data.gcodeIndex = self.data.gcodeIndex + dist
-        try:
-            gCodeLine = self.data.gcode[self.data.gcodeIndex]
-        except IndexError:
-            gCodeLine = 'end of file'
-        print gCodeLine
-        
-        xTarget = 0
-        yTarget = 0
-        
-        x = re.search("X(?=.)([+-]?([0-9]*)(\.([0-9]+))?)", gCodeLine)
-        if x:
-            xTarget = float(x.groups()[0])
-        else:
-            if self.data.units == "INCHES":
-                xTarget = self.gcodecanvas.targetIndicator.pos[0] / 25.4
-            else:
-                xTarget = self.gcodecanvas.targetIndicator.pos[0]              
-        
-        y = re.search("Y(?=.)([+-]?([0-9]*)(\.([0-9]+))?)", gCodeLine)
-        if y:
-            yTarget = float(y.groups()[0])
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Thanks for catching this. I missed it (obviously). 

I agree with @blurfl that it should be clear to the user that they are at one end or the other end of the file, but I think the percent complete indicator can give people that sense because it now stops at 0% and 100%. Mine seems to stop at 99.3% right now, I'll look into that.

Something else that I didn't realize is that by getting rid of the target indicator, you can't see where you are as you fast forward. I found being able to see where I was fast forwarding to to be pretty critical to using the FF effectively so I'm going to add that behavior back in with the regular pos indicator.

I think it would also be cool to be able to press and hold the FF button to continuously FF, but that seems a little tricky so I might hold off on adding that.

Again, thank you @johnbenweber for catching that I broke this when I removed the target indicator. Super important.

BarbourSmith:
I created a new pull request #93 which builds off your work. 

I think for the future, the right way for me to do this would have been to make a PR to your branch which you could accept/reject/add to which we could then merge in in #92 . 

I think it would have been cleaner, because now I've made #92 and #93 which are basically the same thing 💩

blurfl:
The 0% and 100% are enough but having the forward and back buttons grated when they're inappropriate would be very good.

johnbenweber:
Woah, great feedback guys.

@blurfl I agree that some notification might be helpful but your point raises another issue. As you know when stepping through the gcode the lines are printed to the terminal. I don't feel like printing messages (gcode lines included) to the terminal is particularly useful given that GroundControl is likely to be covering the terminal and the user has to somehow discover that this is going on and then switch back and forth. I considered putting the current command somewhere but not sure how to approach it given that the commands can be arbitrarily long.

@BarbourSmith I probably just missed an OBO since the array of commands is zero indexed and compared to the size of the array. I thought I took care of that but can have another look. You might consider drawing the target position only if the index is not equal to the current command index?

@blurfl Disabling the buttons is a good idea which provides feedback to the user in addition to the percentage complete.

BarbourSmith:
I'm hoping that the terminal will go away entirely pretty soon. I removed that the lines are printed to it in #93 because that was mostly in there as a debug thing.

@johnbenweber The missed OBO was me way back when, I just didn't realize it until now 😨

PR #93, opened 2017-03-07
Pr 92 - rev 2

BarbourSmith:
Building on @johnbenweber work in #92 . Added back in the position indicator shown while fast forwarding and fixing an index error so that the % complete ends at 100%.

@blurfl Do you think it's clear enough that the user is at the begning or end of the file? We could disable (turn grey) the buttons to FF or rewind if the user is at the begining or end of the file to make it more clear.

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -114,41 +114,18 @@ class FrontPage(Screen, MakesmithInitFuncs):
         self.shiftY = 0
     
     def moveGcodeIndex(self, dist):
-        self.data.gcodeIndex = self.data.gcodeIndex + dist
-        try:
-            gCodeLine = self.data.gcode[self.data.gcodeIndex]
-        except IndexError:
-            gCodeLine = 'end of file'
-        print gCodeLine
-        
-        xTarget = 0
-        yTarget = 0
-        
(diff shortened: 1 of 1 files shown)


johnbenweber:
I probably should have looked at this before commenting on the other PR.

@blurfl I agree about the behavior of the line number and percentage. We just need to call self.moveGcodeIndex(0) when a new file is loaded.

BarbourSmith:
Great catch @blurfl !

I added a call to moveGocdeIndex which resets it. 

There are still ways to make the system better like holding buttons to FF and disabling the buttons when you are at the edges of the file, but I think this is still much better than it was this morning. 

Any objections to merging this PR as it is and then improving things further in a new PR later?

On a personal note, this is WAY more fun with other people than when I'm the only one looking at this. Thank you two for being involved!

blurfl:
You guys are hot! By all means merge it. 
 Off topic, but any chance of bringing the OSX branch up to date?

johnbenweber:
@BarbourSmith and @blurfl agreed, looks good guys.

BarbourSmith:
Great!

I just brought my build-OSX branch up to date, so it should be good to go.

Is everything still compiling OK?

blurfl:
It compiles 👍, but two issues have surfaced 😔. I’ll open them in that branch, yes?

PR #96, opened 2017-03-09
Workaround for gcode comments being interpreted as coordinates.

johnbenweber:
The target position indicator still has more fun to give. I tried stepping through some gcode which resulted in GroundControl crashing because the gcode had comments containing a 'Y' which it then tried to extract coordinates from.

I would say this is perhaps not an ideal solution but a workaround since the gcode displayed in the view gcode window has all comments stripped and thus does not resemble the original file.

Part of the problem is that the gcode is being parsed in several places so intercepting comments everywhere is cumbersome. I think we need a gcode class that instantiates a data structure on file load and provides a uniform interface to the gcode. It could have get functions for commands, x, y, and z coordinates, feed, etc. For example gcode.command(line) or gcode.x(line) which both ignore comments and a gcode.string(line) which just spits out the line. Thoughts?

### DIFF
--- a/UIElements/viewMenu.py
+++ b/UIElements/viewMenu.py
@@ -67,7 +67,9 @@ class ViewMenu(GridLayout, MakesmithInitFuncs):
         try:
             filterfile = open(filename, 'r')
             rawfilters = filterfile.read()
-            filtersparsed = re.split(r'\s(?=G)|\n|\s(?=g)|\s(?=M)', rawfilters) #splits the gcode into elements to be added to the list
+            filtersparsed = re.sub(r'\(([^)]*)\)','',rawfilters) #removes mach3 style gcode comments
+            filtersparsed = re.sub(r';([^\n]*)\n','',filtersparsed) #removes standard ; intiated gcode comments
+            filtersparsed = re.split(r'\s(?=G)|\n|\s(?=g)|\s(?=M)', filtersparsed) #splits the gcode into elements to be added to the list
             filtersparsed = [x + ' ' for x in filtersparsed] #adds a space to the end of each line
             filtersparsed = [x.lstrip() for x in filtersparsed]
             filtersparsed = [x.replace('X ','X') for x in filtersparsed]


BarbourSmith:
I 100% agree with the plan that we should have a g-code class which handles all of the interpretation and has clear access functions like gcode.x(line). There is no reason that behavior should be duplicated for G1, G0, G2...etc. 

Something to keep an eye on is that some gcode standards omit parts of the line, for example you can see things like:

G01 X10 F123 
Y5 
X12

where the G01 is implied on the second two lines and the federate of 123 is always implied after it is set. For federate, it's almost always set and then implied. Implying the G01 is a little weirder, but you still get it.

I agree that it would be nice to be able to see the comments in the view g-code panel. Comments are sometimes more useful than seeing the raw code because they can include information like that bit size the code was generated for.

That being said, I think having comments masked is better than crashing so I'll merge for now.

Awesome work!

BarbourSmith:
@johnbenweber 

I was checking through the beta tester list and I didn't see you there so I checked the total backer list and I didn't see you there either. Are you under a different name?

johnbenweber:
I can look into writing a class or perhaps check if there is already a python gcode parsing library that I can just wrap around.

No, I'm not a beta tester or backer. I didn't see the project until after the kickstarter ended and couldn't wait. I figured helping out where I can is better than giving you a dollar. To be honest I probably would have built my own anyway because I'm too cheap and I like a good electro-mechanical project. I have comparable hardware all working, just not fully assembled. Hopefully in the next few weeks I'll get a chance to do that.

BarbourSmith:
If you can find a good library we can grab that would be ideal, but if not I think we have most of the elements already to roll our own they just aren't organized into one class.

I was wondering if that was the case. You are making a serous contribution and we appreciate it. I think it's in everyone's best interest that you have working hardware 😀 . If you are happy with the setup you have that's awesome. If you think having the regular kit would be useful, I'm happy to send you a beta kit any time. Just let me know 👍

johnbenweber:
That's a generous offer. Let me see how my setup goes, my biggest concern at the moment is my combination of motor torque and gear teeth but I won't know until I set it up. Rough estimations say it should be fine but at low speed my motors are a bit stally and I'm afraid it will be generally too slow. I'll keep you posted.

BarbourSmith:
Sounds good. Keep us posted.

PR #100, opened 2017-03-13
Fixed z-axis position comparison by considering a "close enough" tolerance.

johnbenweber:
Comparing floats using the '==' operator is subject to roundoff error and was causing erratic line color/type when displaying the cut path on the canvas. I added a tolerance which I assume is smaller than that physically achievable. See [link](http://stackoverflow.com/questions/5595425/what-is-the-best-way-to-compare-floats-for-almost-equality-in-python) for some spirited general discussion.


BarbourSmith:
👌 Congratulations on PR 100 🎉 🎈 💥

johnbenweber:
Congratulations indeed. I have more coming.

PR #115, opened 2017-03-20
patch otherFeatures for serial port issue

blurfl:
Upon activating the Actions tab, a call to self.connectmenu.updatePorts() found in UIElements/otherFeatures.py breaks the serial connection, at least under OSX. Commenting out the offending line is a bandaid.


BarbourSmith:
I'm seeing no line changes between this version and master, is that right?

blurfl:
Oops😟 I'll fix it in a bit

blurfl:
Ok, there it is. ☺️

BarbourSmith:
On windows it seems to be listing all of the possible COM ports, not just the ones connected which makes sense from looking at the code which is `sysports = ['COM%s' % (i + 1) for i in range(256)]`
all ports

Is there a way around that? Could we just use the old method, but only on windows?

davidelang:
It's not possible to know if a serial port is connected unless you try sending 
data to it (and if that port is used by something else, doing so will corrupt 
dtat for that device)

BarbourSmith:
What if we replaced the line `sysports = ['COM%s' % (i + 1) for i in range(256)]` with the old version `sysports = self.data.serialPort.listSerialPorts()`. 

We would still be touching the serial port, but we could check to see if we're already connected by checking the flag self.data.connectionStatus in the file otherFeatures.py. If the firmware is already connected we could skip updating the ports, if it is not connected we could check the ports without worrying about ruining the data because there won't be any.

blurfl:
Linux (kivypie) doesn't seem happy either. I'll try special-casing 'darwin' only.

BarbourSmith:
@blurfl good catch testing on kivypie

blurfl:
Closing this PR, to open a new one with a darwin-only patch

PR #116, opened 2017-03-21
patch-only-darwin

blurfl:
Win and linux work properly with previous approach, patch only the case
where platform is darwin


BarbourSmith:
Looks great!

BarbourSmith:
I'm getting an error:

`File "C:\Users\Bar\Git\GroundControl\UIElements\screenControls.py", line 18, in show_actions
     content.setUpData(self.data)
   File "C:\Users\Bar\Git\GroundControl\UIElements\otherFeatures.py", line 15, in setUpData
     self.connectmenu.updatePorts()
   File "C:\Users\Bar\Git\GroundControl\UIElements\connectMenu.py", line 31, in updatePorts
     portsList.append(port)
 UnboundLocalError: local variable 'port' referenced before assignment`

I think it's just a typo in line 31 `for ports in self.data.serialPort.listSerialPorts():` where `for ports` should be `for port` singular

blurfl:
It test OK on windows, but the Pi isn't happy. Let me work on it some more.

blurfl:
Need to handle linux as well. Closing, will open another PR

blurfl:
Yes, thanks. I've fixed that and reworked the Linux  case, in a new PR. 
I noticed that in Win, the port currently connected does not appear on the list; is that the desired behavior? Hope so, don't know how to change it 😄

PR #117, opened 2017-03-21
patch connectMenu for serial port issue

blurfl:
Upon activating the Actions tab, a call to
self.connectmenu.updatePorts() found in UIElements/otherFeatures.py
breaks the serial connection, under OSX and linux.


blurfl:
This handles OS X and linux port lists in the expected manner and does not break the serial connection if already established. The win case is left as before - this shows ports except for the port currently connected, on WIN10 at least, and the serial connection is preserved. Someone to verify that my assessment of how Windows works is correct.

BarbourSmith:
Looks great to me!

I tested it on windows and everything is 👌 

You are right that on windows the connected port doesn't show. I'm not sure that's the desired behavior, but I don't think it's a big problem either.

PR #120, opened 2017-03-22
Fix circles in haas generic

BarbourSmith:
A better version of the parser to catch HAAS-generic code, cleans up the way gcode is parsed and processed, needs to be tested against firmware throughly because a lot less processing is done on the Ground Control side now to clean up the code, which feels like the right thing long term, but might break things short term


BarbourSmith:
I'm going to close this one unmerged- I think it needs more time to not break everything than I have to spend on it right now

Issue #73, opened 2017-02-09
Improved Calibration Routine


BarbourSmith:
I've been talking to Keith Selbo (our resident mathematician) about how to make the calibration of the machine as easy as possible and he made some EXCELLENT suggestions that I wanted to get written down.

The problem we are solving is that the machine needs to be calibrated in-order to work accurately. Because of the non-linear way in which the machine relates rotational motion of the motors to movement it is important that the machine know the distance between the motors and amount of chain paid out. These values are stored in nonvolatile memory so even if power is lost the machine won't loose it's calibration.

Measuring these dimensions with a tape measure is going to result in user error and confusion. Keith suggested that the machine has a built in measuring system which would be more accurate than a tape measure. Basically we can use the motors, encoders and chain to establish the machine's dimensions.

Here's my slightly modified version of how this could work.

# To measure the distance between the motors accurately:
1) Set both motors to have one gear tooth pointing straight up
2) Attach last link on left chain to the top tooth on the left motor
3) Pay out chain until it reaches the right motor, hook chain onto top tooth
4) Reverse left motor direction to tension chain removing droop. The tension can set automatically because the motor voltage and speed of rotation are known. If we need to we can take into account chain droop but I don't think it will be significant
5) The value of the left motor encoder will now reflect the distance between the motors

# To measure the length of chains
1) Set both motors to have one gear tooth pointed straight up
2) Use the motors/encoders to pay out the correct length of chain. This will have to be done one chain at a time because until there is some tension on the chain the sprockets are not self tailing and can catch the chain.

BarbourSmith:
Added

PR #122, opened 2017-03-24
Switch to measuring dist between motors

BarbourSmith:
Switch to measuring the distance between the motors in settings instead of the motor offset distance. 

The distance between the motors is easier to measure and makes more sense to reduce the chance of confusion.

This pull request needs to be merged at the same time as a similar pull request in firmware

### DIFF
--- a/main.py
+++ b/main.py
@@ -75,10 +75,10 @@ class GroundControlApp(App):
         },
         {
             "type": "string",
-            "title": "Motor Offset Horizontal in MM",
-            "desc": "The horizontal distance from the corner of the work area to the motor.",
+            "title": "Distance Between Motors",
+            "desc": "The horizontal distance between the center of the motor shafts in MM.",
             "section": "Maslow Settings",
-            "key": "motorOffsetX"
+            "key": "motorSpacingX"
         },
         {
             "type": "string",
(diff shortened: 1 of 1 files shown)


johnbenweber:
I'm not sure how the kinematics works now but there used to be a constant for the diameter of the gear on the motor in the firmware. My gears are a different size than the stock ones so it would be nice if I didn't have to change the firmware. Same for encoder steps per revolution. Perhaps there could be an advanced section in the ground control settiings?

davidelang:
If there is a constant/variable for the diameter of the gear, it should be 
changed to the number of teeth in the gear and the link size. People can easily 
count the number of teeth, and from that and the link size, the diameter can 
easily be calculated. But getting someone to properly measure the diameter of a 
gear is hard, and a small error in the diameter will translate to a large error 
later on.

davidelang:
also see #73 and #127 This should be measured, but not by the builder, it should be measured by the machine

johnbenweber:
I agree that the number of teeth and chain pitch is an easy way to work out the pitch diameter of the gear. I've seen the algorithm for determining distance between the motor centers but I think number of gear teeth/chain pitch is something you'll have to input if you have a non-standard setup. I'm not sure how the machine would work that one out without some sort of limit switch setup which I feel adds unnecessary complication. Looks like advanced settings is my next project.

BarbourSmith:
👍 👍 👍 

@davidelang I agree 100% that we really want the machine to be measuring itself. This PR is one step in that direction (ie using the distance between the motors which the machine will measure itself instead of some arbitrary point (the corner of the plywood) which would be hard to reference). Your comments in #127 are spot on.

@johnbenweber You are looking for line 45 in CNC_Functions.h which is 

```python
#define DIST_PER_ROTATION 10*6.35//#teeth*pitch of chain
```

How's everything going with the build? Anything we can do to help?

PR #128, opened 2017-03-25
Moved the onscreen buttons so that they are more grouped by function …

johnbenweber:
…and capitalized the stop button for consistency. Top image is before changes and bottom is after.
before
after

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -50,7 +50,7 @@
         id: gcodecanvas
     BoxLayout:
         orientation: 'vertical'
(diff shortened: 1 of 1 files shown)


BarbourSmith:
👍 Looks fantastic! I'll test it real quick and then merge!

Thanks for showing the side by side pictures 👍

BarbourSmith:
Looks great. I'm getting some weird behavior on my z-axis which I think has to do with the fact that clicking the arrow buttons sends commands with a federate of 0, but that's the same on master.

Thanks for doing that!

PR #129, opened 2017-03-25
Added advanced settings to correspond to those implemented in Firmware.

johnbenweber:
As with Firmware pull request [143](https://github.com/MaslowCNC/Firmware/pull/143) I haven't actually thoroughly tested the changes. Have a look and I can fix any lingering issues.

### DIFF
--- a/main.py
+++ b/main.py
@@ -107,20 +107,46 @@ class GroundControlApp(App):
             "desc": "The path to the open file",
             "section": "Maslow Settings",
             "key": "openFile"
+        }
+    ]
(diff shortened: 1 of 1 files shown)


BarbourSmith:
..And you've already one what I was asking about in my other comment. AWESOME.

I'll test both on the machine and then merge

BarbourSmith:
Looks good. I haven't tested the firmware side yet, but I think it's safe to merge these and then test because unread values are just ignored on the firmware side

Issue #125, opened 2017-03-24
Add a 'Stop' or 'Pause' tutor to the pop up; currently the user can only 'Continue'. 


blurfl:

BarbourSmith:
👍

BarbourSmith:
added in #136

PR #136, opened 2017-03-26
Add hold button to dialog

BarbourSmith:
The dialog that prompts the user for action if the USB connection is lost mid cut or if the machine needs other attention used to only have the option to "Continue". This PR adds a second button which will command the machine to "Hold" so the user can take approprate action. Addresses issue #125

### DIFF
--- a/UIElements/notificationPopup.py
+++ b/UIElements/notificationPopup.py
@@ -12,5 +12,6 @@ class NotificationPopup(FloatLayout):
     If the text does not all fit on the display, the user can scroll it
     
     '''
-    cancel = ObjectProperty(None)
+    continueOn = ObjectProperty(None)
+    hold = ObjectProperty(None)
     text = StringProperty("")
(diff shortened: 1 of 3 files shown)


BarbourSmith:
@blurfl When you get a chance take a look at this one and see if adds all the behavior you were looking for in #125 . 

Great suggestion 👍 👍

blurfl:
?Yes, this is the idea. Back on the front page there is nothing that indicates we’re ‘on hold’, but that can be another issue for a different day. Good work!

BarbourSmith:
@blurfl Great point about the need for a "we're on hold indicator".

What are you thinking would be ideal? What if we changed the text of the "Run" button from "Run" to "Resume"?

blurfl:
Yes, that would show ‘where to click’ to get going again. How hard is that?

BarbourSmith:
I don't think it should be too hard. Kivy has this build in event driven framework which will update everything automatically. Basically if the text is of type `StringProperty` then all we have to do is change the string and the button will refresh itself. It's slick.

BarbourSmith:
Will you file a new issue for it? I'm going to merge this one as is.

blurfl:
OK, I’ve opened #143 for it.

PR #141, opened 2017-03-27
Added about popup and corresponding text.

johnbenweber:
Feel free to modify the actual text content, I'm not sure what exactly to say.


BarbourSmith:
Looks great! We've got to bump the version number up from v0.60 to v0.63, but I don't think that's an issue for this pull request.

Great work!

PR #142, opened 2017-03-27
Firmware version on startup

johnbenweber:
Added function and corresponding call to send B05 to machine before sending G20/21 on connection start. This should cause the firmware version to be printed in the text console but I haven't actually tested it yet.


BarbourSmith:
Perfect. 

Looks like the firmware version number needs to be bumped too. Now that we have ways to see these things, I think I will have to do a better job of making sure they are correct.

Fantastic work. Thank you. 👍 👍 👍

PR #155, opened 2017-03-29
Make 'HOME' button lift z-axis then move home

BarbourSmith:
Zaxis will lift, move to home, then lower z-axis to zero. 

Needs to be tested with both inches and mm set before merge


blurfl:
This works correctly on my system, MM and INCHES.

PR #156, opened 2017-03-29
Clean up data structure

BarbourSmith:


BarbourSmith:
As far as I can tell this doesn't break anything, but I'm going to leave the junk commented out rather than deleted for a week or so to see if anything goes wrong

davidelang:
just a reminder that since you are using git, you can pull any contents that 
have been in the file and re-include them (several different ways, including 
just doing git revert on the commit that removed them)

It doesn't hurt to leave them commented out for a bit, but it's not neccessary.

BarbourSmith:
Great point!

PR #162, opened 2017-03-30
readme.txt

TheRiflesSpiral:
A generic storage location for images used during install.


TheRiflesSpiral:
I couldn't figure out how to add images to the folder without the Images folder being created first. Once the folder is there I'll upload the images and update the instruction with the commands to get them into Kivy properly.

Issue #165, opened 2017-03-31
Empty increment size crashes GroundControl


GeroBH:
On
Ubuntu 14.04.5 LTS trusty
Python 2.7.6
Dragon.nc loaded
not connected

Clicking "increment size of movement" opens with blank value, no cancel button, clicking "done" without entering anything replaces current increment with "blank". Clicking on "units" afterwards makes Ground Control crash.

File "/home/ghadmin/Maslow/GroundControl/GroundControl-master/UIElements/frontPage.py", line 94, in onUnitsSwitch
     self.moveDistInput.text = str(float(self.moveDistInput.text)/25)
 ValueError: could not convert string to float:

BarbourSmith:
I can replicate this on windows. 

Thank you for spotting that. Should be a pretty easy fix. 

I think the real question is what is the right behavior if you click "done" without typing anything. I'd say we should just leave the value what it was before the popup was opened.

BarbourSmith:
Fixed in #165

PR #168, opened 2017-04-01
Entering invalid dist to move keeps current value

BarbourSmith:
Fix for #165

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -557,49 +557,48 @@
 
 <TouchNumberInput>:
     textInput:textInput
+    cols: 1
+    size: root.size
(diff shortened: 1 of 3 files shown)


groundcontrol.kv
```diff
@@ -557,49 +557,48 @@
...
-                text: "."
-                on_release: root.addText(".")
-            Button:
-                text: "0"
-                on_release: root.addText("0")
-            Button:
-                text: "Done"
-                on_release: root.done()
\ No newline at end of file
+        cols: 3
```
BarbourSmith: This change is unrelated to the issue, but fixes a warning that was showing up when the popup was created

Issue #157, opened 2017-03-29
make this gcode run in GC


hannahebteagle:
hello! :) attached is the ZIPPED version of the .nc that won't open correctly yet in Ground Control :)))
[untitled folder 2.zip](https://github.com/MaslowCNC/GroundControl/files/880214/untitled.folder.2.zip)

BarbourSmith:
Fixed in #170

BarbourSmith:
There appear to be some scaling issues with the files, but that's not a Ground Control issue

BarbourSmith:
The latest changes appear to have made this code not work again, more investigation is needed

BarbourSmith:
First thing is this file doesn't specify units (G20/G21) or coordinate system (G90/G91) at the beginning which is poor form. Ground Control guesses that the units are the same as the last file opened, which isn't ideal, but there's not a lot else we can do about it.

BarbourSmith:
Fixed again in PR #193

PR #170, opened 2017-04-01
Add support for NXXX appearing before gcode

BarbourSmith:
It's a pretty common thing for gcode to have the prefix N### to indicate the gcode section number so this PR adds support for it rendering in GC. 

This needs to be tested in conjunction with the frimware before it is pulled in.


BarbourSmith:
Ground Control side fix for issue #157

PR #186, opened 2017-04-10
Fix for issue 176

BarbourSmith:
The gcode canvas now loads further zoomed out.


UIElements/gcodeCanvas.py
```diff
@@ -83,8 +83,13 @@ def reloadGcode(self, *args):
             self.data.gcodeFile = ""
         
     def centerCanvas(self, *args):
+        print "center canvas"
```
BarbourSmith: 🎋

Issue #160, opened 2017-03-30
Rasp Pi Touchscreen keyboard types double characters


blurfl:
Using a keyboard to enter values in the settings produces double characters per keystroke. Typing  a '1' produces '11'. This makes entering values impractical.

TheRiflesSpiral:
Check the config.ini file for multiple or generic HID input types.

[input]
mouse = mouse
mtdev_%(name)s=probesysfs,provider=mtdev
**hid_%(name)s=probesysfs,provider=hidinput
%(name)s=probesysfs,provider=hidinput**

The last two lines in this file cause USB HID input to be handled by two providers. Try deleting the last line.

blurfl:
Once again, you've done it. 👍 Thanks for your help.
For posterity, on my (generic) kivypie setup. I edited 
/home/sysop/.kivy/config.ini and commented out the line:
%(name)s=probesysfs,provider=hidinput

Note that this change also corrected an issue with mouse-clicking which affected the mouse but not the touchscreen, causing double-clicks instead of single clicks.

TheRiflesSpiral:
I'll update the instructions to include this edit in KivyPi.

BarbourSmith:
Beautiful work everyone!

BarbourSmith:
Has this issue been resolved?

blurfl:
I remember some discussion about this after @theRifleSpiral posted the fix on the forum.I bumped the thread to see whether anyone thinks it should stay open. It certainly worked for me, “close” is my vote.

TheRiflesSpiral:
If you're running python under root (sudo python main.py from the ~/GroundControl directory) Kivy will use the root's config.ini file, not sysop. I updated the wiki to include an edit to both the sysop and root users.

BarbourSmith:
Fantastic! Thank you for updating the wiki to include both edits, and good sleuthing to figure out what was going on!

PR #190, opened 2017-04-11
Fix for firmware issue #160

BarbourSmith:
When clicking the "home" button the machine asks the user to adjust the
z-axis but does not wait for the adjustment to take place as described
in firmware issue
[160](https://github.com/MaslowCNC/Firmware/issues/160) . This fixes the
problem by asssuming manual control over the z-axis if there is no
z-axis attached.

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -249,14 +249,23 @@ class FrontPage(Screen, MakesmithInitFuncs):
         
     def home(self):
         
-        if self.units == "INCHES":
-            self.data.gcode_queue.put("G00 Z.25 ")
+        print "This!"
+        print self.data.config.get('Maslow Settings', 'zAxis')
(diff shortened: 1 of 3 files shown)


UIElements/frontPage.py
```diff
@@ -249,14 +249,23 @@ def zeroZ(self):
...
     def home(self):
         
-        if self.units == "INCHES":
-            self.data.gcode_queue.put("G00 Z.25 ")
-        else:
-            self.data.gcode_queue.put("G00 Z5.0 ")
-        
-        self.data.gcode_queue.put("G00 X" + str(self.data.gcodeShift[0]) + " Y" + str(self.data.gcodeShift[1]) + " ")
+        print "This!"
+        print self.data.config.get('Maslow Settings', 'zAxis')
```
BarbourSmith: ⚜️

Issue #206, opened 2017-04-17
z-axis - Menu just for it


barryhund:
Please give Zaxis its own menu for calibration independent of sled movement distance setting.

BarbourSmith:
Added in PR #208 .

If you are running GC from the source, take a look and let me know what you think?

Is it missing anything?

BarbourSmith:
How do we feel about closing the issue?

Issue #154, opened 2017-03-29
Separate the distance setting for Z-axis move amount from the horizontal distance setting


blurfl:
There is some interaction between the 'Distance to Move' setting and the distance the Z-axis motor travels for the z-in or z-out button. The relationship is not clear, and when large distances are set for horizontal traverse, the z travel will be large as well. It is possible to run the z-axis mechanism into its physical limits causing damage to the equipment.
 Having a separate setting for z-axis distance would solve the issue. It would not need to be as accessible as the horizontal distance setting, perhaps it could live on the settings page rather than on the front page.

BarbourSmith:
Good idea. Right now the relationship is 10:1 so the z-axis will move 1/10th as far as the XY axis. 

I think figuring out where to put that dialog is going to be the tricky part. What about replacing the the unused "test" button with a z-dist?

blurfl:
I understand now. I've been using large values to move the sled around the workspace, and when I go "Home" the automatic z kicks in. I'll unlatch the router before traversing.
 The "test" button would be fine, though putting it on the settings page would do too. I suppose it depends on how often one would use the 
'Test" function. I use it (from Actions) pretty often, trying to dial in some accuracy 😕 in the vertical direction.

BarbourSmith:
I broke everything related to the z-axis out into a popup. 

What do you think? Is it better? Am I missing anything?

blurfl:
I’ll try it out - too sounds like a good idea, though.

BarbourSmith:
Thanks, it wasn't really my idea tho :)

blurfl:
Those are all the z-things, conveniently grouped! Not sure how much you want tested, but here goes :)

- It takes a bit of figuring out - inches/mm shows the current setting, but switching between the two doesn't change the distance value as it does for the xy setting. Could changing the units keep the setting and show the value in the new units? 
-  I'm only able to use the 'in' and 'out' buttons once; after that the z-motor whines but does not move.
-  No matter what distance value I choose (.25 inches for instance) if I leave the popup with 'done' and come back the distance value is set to .1 ; 
-  Could 'In' and 'Out' be 'Z-In' and 'Z-Out' to avoid confusion about in and inches?

BarbourSmith:
Those are all fantastic points which are easily addressed. Thanks for the guidance!

BarbourSmith:
I've pushed a new version with the following changes:

[...]

PR #208, opened 2017-04-18
Fix for 154

BarbourSmith:
Break the z-axis controls out into a popup instead of having them shared with the XY axies


UIElements/zAxisPopupContent.py
```diff
@@ -0,0 +1,78 @@
...
+        
+        self.unitsBtn.text = self.data.units
+    
+    def zIn(self):
+        '''
+        
+        Move the z-axis in
+        
+        '''
+        print self.distBtn.text
```
BarbourSmith: ⛲️

BarbourSmith:
Address the issues #154 and #206

PR #220, opened 2017-04-22
Return to center lifts z

blurfl:
"Return to Center" should lift z then traverse and finally zero z.

### DIFF
--- a/UIElements/runMenu.py
+++ b/UIElements/runMenu.py
@@ -12,6 +12,6 @@ class RunMenu(FloatLayout):
         App.get_running_app().stop()
     
     def returnToCenter(self):
-        self.data.gcode_queue.put("G00 Z0 ")
+        self.data.gcode_queue.put("G00 Z0.25 ")
         self.data.gcode_queue.put("G00 X0 Y0 Z0 ")
         self.parentWidget.close()
--- a/UIElements/runMenu.py
+++ b/UIElements/runMenu.py
@@ -12,6 +12,6 @@ class RunMenu(FloatLayout):
         App.get_running_app().stop()
     
     def returnToCenter(self):
-        self.data.gcode_queue.put("G00 Z0.25 ")
+        self.data.gcode_queue.put("G00 Z0 ")
         self.data.gcode_queue.put("G00 X0 Y0 Z0 ")
         self.parentWidget.close()
--- a/UIElements/runMenu.py
+++ b/UIElements/runMenu.py
@@ -13,5 +13,6 @@ class RunMenu(FloatLayout):
     
     def returnToCenter(self):
         self.data.gcode_queue.put("G00 Z0.25 ")
-        self.data.gcode_queue.put("G00 X0 Y0 Z0 ")
+        self.data.gcode_queue.put("G00 X0 Y0 ")
+        self.data.gcode_queue.put("G00 Z0 ")
         self.parentWidget.close()


BarbourSmith:
Looks great! Thank you for the pull request!

BarbourSmith:
I had two issues with similar behavior for the "Home" button that might come up here.

1) Sending all three commands at once works great if the user has a z-axis installed, because each command is executed one right after the other. If there isn't a z-axis installed the machine still executes each command one right after the other, but there is no pause to let the user respond. Last time, I worked around this issue by simply moving directly home if there is no z-axis installed.

2) If the user is working in mm raising to a height of .25mm is probably not enough to clear the wood

I ended up doing it like this:

```python

#if the machine has a z-axis lift it then go home
        if int(self.data.config.get('Maslow Settings', 'zAxis')):
            if self.units == "INCHES":
                self.data.gcode_queue.put("G00 Z.25 ")
            else:
                self.data.gcode_queue.put("G00 Z5.0 ")
            
            self.data.gcode_queue.put("G00 X" + str(self.data.gcodeShift[0]) + " Y" + str(self.data.gcodeShift[1]) + " ")
            
            self.data.gcode_queue.put("G00 Z0 ")
        #if the machine does not have a z-axis, just go home
        else:
            self.data.gcode_queue.put("G00 X" + str(self.data.gcodeShift[0]) + " Y" + str(self.data.gcodeShift[1]) + " ")
```

What do you think?

blurfl:
That makes sense, much more sophisticated.

BarbourSmith:
I did it the other way first 😀

BarbourSmith:
Ignore the `str(self.data.gcodeShift[0])` that's to take it to the "Home" location, in this case we can just use 0

blurfl:
'self.units' need to be 'self.data.units' and works as hoped except if I leave the sled raised and off-center and quit GC. When I start GC again and click "Ret..Center" the z-axis begins moving back in before the sled gets to 0,0. Once I've "Ret..Center" the first time, it works as hoped the rest of the session.
When GC starts up, the z position from last session is shown, but the xy values are 0. That's probably why the z begins moving while the sled is traversing this first time.

BarbourSmith:
Good catch on that corner case!

blurfl:
I'm not sure there's any way to fix that corner case other than leaving the z axis alone. Any ideas? You're busy with much more important things, but if you have a suggestion I could look farther into this.

BarbourSmith:
I think your description of what is happening is spot on. The real solution is that the machine needs to remember it's XY position when it powers off, or we need to be able to recreate the XY position from the chain lengths. That's a separate issue.

I'd say throw in the z-axis stuff as you have it working with self.units substituted for self.data.units and lets merge it!

blurfl:
Bar,
 I’d like you to rescind my direct access to GroundControl; there’s something basic about GitHub Desktop that I haven’t mastered yet, and I seem to keep committing the wrong versions of things. I’m sorry for the confusion I’m causing, I very much fear causing real damage to the project.

BarbourSmith:
Hahah, you got it! :-)

From my end it looks like you are doing nothing but good 💯

BarbourSmith:
You are back to regular access level. As always, thank you for everything you do!

blurfl:
Thanks, I feel safer now. I saw those 24 changes and knew I was out of my depth.

PR #221, opened 2017-04-22
Change the label on "Define Origin" to "Set Home"

BarbourSmith:
This is the place the machine moves back to when you click the "Home"
button so it makes sense that it would be called "Set Home"

Anyone oposed?

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -124,7 +124,7 @@
                 text: root.units
                 on_press: root.switchUnits()
             Button:
-                text: 'Define\n  Home'
+                text: 'Define\nHome'
                 on_release: root.moveOrigin()
                 disabled: False
         GridLayout:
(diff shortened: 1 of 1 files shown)


blurfl:
"Set Home" makes sense to me: 
My understanding is that "Origin" refers to the point arrived at with "Return to Center", the navel of the world, from which all calculations are reckoned. "Home" is the starting point from which a gcode file makes relative movements. GC takes care of mapping the "Home" coordinates into "Origin" coordinates.
 "Origin" is set by calculations from the settings used by "Calculate Chain...", changed rarely, when calibrating the machine.
 "Home" is determined by moving the sled and clicking "Set Home" and is changed frequently, when moving to a spot on the work area for the next cut.

BarbourSmith:
Great concise description of what is happening. I feel like that description belongs on the wiki somewhere.

Also, the phrase "naval of the world" made my day :-)

PR #228, opened 2017-04-24
Make gost position indicator show up again when using FF

BarbourSmith:


blsteinhauer88:
Thanks bar it worked great

Issue #233, opened 2017-04-29
Lost connection and GC reconneting behaviour


GeroBH:
Sorry I can't come up with a better title, feel free to change it.
Also explaining it, is a bit difficult.
But what I have observed is a potential danger. I'll try.

After a 'buffer overflow' and a ' USB connection lost' popup, with out any response buy the user, the sled moves downward. I assume it moves to where GC thinks the centre is. However, that is off after loosing connection, and a cutting bit should not move at all, after that, in my opinion. Just freeze.
The 'continue' button as also something I would like to question.
I my case, cutting far up on the sheet, after the 'buffer overflow' and a ' USB connection lost', I clicked 'hold'. GC had reconnected and had set its router position far lower then it was. There sled was high and I wanted to bring it down, to get the chains off a recalibrate the system. Clicking the 'down arrow' moved the sled even more up, while the symbol for the sled went downwards on the screen.
I would love to try the same on the lower part of the sheet, but the attempt of GC to lift my sled above the sheet, seem to have fried the motor controller. And the right one, resulting that Z is also not functional any more.
What is the right behaviour after a lost connection? Freeze and insist on a new calibration?

jbarchuk:
Problem response. This is a MODERATELY IMPORTANT SAFETY ISSUE. Something
along the lines of what I write here should probably be added to the user
guide.

Gero, your idea that something needs to be done better after a
communications failure is of course correct, but that's not what I'm
writing about.

> After a 'buffer overflow' and a ' USB connection lost' popup, with out
any response buy the user, the sled moves downward.

That's what I call a 'spooky' bug. Meaning with some failure, the machine
continues into an unknown and probably uncontrolled action.

> I clicked 'hold'. ...
> Clicking the 'down arrow' moved the sled even more up, while the symbol
for the sled went downwards on the screen.

What you did after that problem was incorrect, but you didn't know that.
When plain vanilla software crashes, about the worst result is lost data.
When hardware is involved, the possibilities of what a machine -might- do
is limited only by its speed and strength. Maslow is very slow (right now)
but also very powerful. In this case the sled, chains and motors
can't go very far, or too fast, but the motors can definitely break the
frame, and that could even include flying splinters or whipped chains.
Breaking the motor mount actually isn't a tragedy. If it fails the motor
and chain drop to the floor. It's the weakest point, and making a new mount
isn't the worst price of failure. However some folks are making different
stronger mounts, and the next weakest point might be the chain-sled
[...]

PR #234, opened 2017-04-29
Make USB connection lost message more serous

BarbourSmith:
The message associated with the USB connection dropping did not empizise
the importance of what happened.


BarbourSmith:
Partial fix for #233

PR #265, opened 2017-05-06
Added filter list to cover most common gcode file extensions.

johnbenweber:
Added a filter list to the kivy filechooser object based on this list of gcode file extensions [link](http://bobcadsupport.com/helpdesk/index.php?/Knowledgebase/Article/View/13/5/known-g-code-file-extensions). They are easy to add if you think of more.

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -480,7 +480,7 @@
         FileChooser:
             id: filechooser
             path: root.path
-            filters: ['*.apt','*.cnc','*.din','*.dnc','*.ecs','*.eia','*.fan','*.fnc','*.fgc','*.gc','*.gcd','*.h','*.hnc','*.i','*.m','*.min','*.mpf','*.mpr','*.nc','*.NC','*.ncc','*.ncg','*.ncp','*.ngc','*.plt','*.ply','*.rol','*.sbp','*.tap','*.xpi']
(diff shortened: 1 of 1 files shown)


johnbenweber:
Whoops, I see there is already a valid extensions setting in GroundControl. I'll change the filter to only include those since otherwise GroundControl will complain when you try to open other ones.

BarbourSmith:
Looks fantastic! Great idea. Thank you so much for all these fantastic pull requests.

The settings option to limit the file types doesn't do anything really anymore right? Should we just get rid of it?

johnbenweber:
Well, I think the validExtensions key will make Kivy throw an error if you pick a file type not in the list. Since the files are being filtered you can't choose files which don't those types so I suppose it no longer does anything. It's probably not critical if a user ever did somehow pick another file format anyway.

It might be nice to parse the validExtensions key (which looks like `'.txt, .exe, .etc'`) into an array (which looks like `['*.txt','*.exe','*.etc']`) to use as the filter so it isn't buried deep within the .kv file but I dunno if that's worth the effort.

I know I looked into a fancy filechooser before, I'll have to look around to see if I can find it again. It would be nice to have the standard pull down that allows you to choose filtered extensions or all files among other things.

PR #266, opened 2017-05-06
Grey out buttons when no connection

johnbenweber:
Greys out all of the buttons I could find which make the machine move when the connection status flag is false. I'm not sure whether the machine settings should be greyed out as well but that's easily done if necessary.

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -57,6 +57,7 @@
             size_hint: None, None
             width:dp(300)
             height: dp(60*4)
+            disabled: not app.data.connectionStatus
             Button:
                 on_press: root.upLeft()
                 canvas.after:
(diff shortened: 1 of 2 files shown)


BarbourSmith:
Looks perfect! AWESOME! 👍 👍 👍 💯 💯 💯

BarbourSmith:
I'm seeing some strange behavior where sometimes not all the buttons come back from being disabled. They all work normally, but some of them still look disabled.

Anyone have any idea what could cause this?

buttons look funny

johnbenweber:
Well that's not ideal. Interesting because the layout for the jog buttons is disabled as a whole. Not sure why individual ones would stay greyed out. It might be caused by the fact that the old z-in and z-out buttons were individually disabled.

PR #268, opened 2017-05-06
Page through lines in View Gcode popup

johnbenweber:
Added buttons and ability to page through lines 447 at a time. Also made the popup title more informative. This will probably conflict with PR #267 but I can easily resolve that.


BarbourSmith:
Looks fantastic! 👍 👍 👍 👍 👍 👍

PR #271, opened 2017-05-07
Removed individually disabled buttons which used to be zin and zout.

johnbenweber:


BarbourSmith:
That's a good idea.

I tested it some more and couldn't get the grey thing to happen again, but it only ever happened sometimes before so let's keep an eye on it.

For the record, I think the way you implemented the whole thing looks perfect...this is def flaky behavior on Kivy's part.

Great idea for something that might make it happier.

Issue #264, opened 2017-05-06
View Gcode and scrolling zooms out


GeroBH:
If you use View Gcode and scroll though a long file, if you don't scroll back up, you can't see you worksheet any more. You have to zoom back in, the same amount that you have scrolled down in the Gcode. Not a big issue, but worth mentioning. I got confused for a second.

johnbenweber:
Yeah, would be good to fix this. This is mostly why I implemented the "reset view" button under the "view gcode" button.

johnbenweber:
Should be fixed, see PR #280.

PR #280, opened 2017-05-14
Changed canvas zooming behavior to prevent unintentional zooming.

johnbenweber:
See issue #264. The `on_motion` event was bound to the window object and captures global key presses and was causing zooming of the gcode canvas when other popups were in focus. The `on_touch_up` event only fires when the gcode canvas is in focus.


BarbourSmith:
Fantastic!

Issue #225, opened 2017-04-23
Transport buttons not responding


nwrancher:
I am not able to use the transport < << >> > buttons for any useful function.  Sometimes they appear to make it skip ahead, but I don't understand how to manage it so it isn't dragging the bit.

I would like to see them greyed out when code is running, active when on hold or no code running.  Ideally they would scroll the code in the viewer with the active line highlighted and position indicator showing where on the cut, then "enter" to run starting on that line.  Or something.  I'm not entirely sure how they should work, but currently, for me, they don't seem to.

BarbourSmith:
Yes, I agree those buttons have been neglected while everything else changed and they don't really work anymore. We need to fix them.

How they work right now is that they jump ahead to either 1> or 10> lines of gcode. None of the skipped lines are run, so if the bit is supposed to retract during that skipped part it doesn't. There used to be an indicator of where the machine was going to skip to, but that doesn't show up anymore as described in issue #227 

What is the behavior we want to see when clicking the arrow buttons? Do we want it to run the intermediate lines (that wouldn't be much of a fast forward)?

It shouldn't be too hard to fix, but I'm not clear right now on what we want them to do.

blurfl:
I think it is more useful to skip lines, that's the only way we have to ignore part of a file and move on. I suppose adding a 'Skip/Run' button would offer both options, but more buttons :( ?

BarbourSmith:
That makes sense. I think the first step is to put back in the indicator of where we are skipping to as described in #227 . Then maybe we can get a better sense of what we want.

blurfl:
👍

jbarchuk:
More options = less restrictions. A user might want to review steps ahead,
go back to the point where it stopped, and continue from there. For example
'I'm not exactly sure where the router's gonna go from here, so I want to
check the path first. At this point I have the options of quitting, going
back/forth manually in the script (a shuttle in media editing terms) and
continuing forward from that point, or going back to where the GC remembers
the last point where the script was stopped.' When the user has been
thinking about things and moving the sled around, it's less error prone for
the GC to remember where it stopped rather than the user.

johnbenweber:
I think it might be more useful to skip to the next z-move than a specific number of line. When I've wanted to skip it was to skip over a whole feature or a depth pass and you don't necessarily know how many lines that is.

blurfl:
I've often used the step over lines, but also wished for a 'skip to next z'... Another button!

[...]

PR #282, opened 2017-05-17
Skip through Gcode by Z move

johnbenweber:
See discussion in issue #225. This seems to work in ground control but I haven't actually tested it with hardware. Added a function `isClose` for comparing positions with a tolerance that changes based on unit system. The default tolerances may need to be adjusted. The logic in the `moveGcodeZ` could probably be improved.

### DIFF
--- a/DataStructures/data.py
+++ b/DataStructures/data.py
@@ -32,6 +32,8 @@ class Data(EventDispatcher):
     comport    = StringProperty("")
     #The index of the next unread line of Gcode
     gcodeIndex = NumericProperty(0)
+    #Index of changes in z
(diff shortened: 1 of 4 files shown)


BarbourSmith:
Very cool!

Can anyone verify that this lets you jump to the next z-axis move and test it in hardware?

 I won't have a machine for next few days because of Maker Faire

johnbenweber:
Be careful with this, I think I just fixed an obo issue.

BarbourSmith:
I think this looks fantastic. 

Great work! 👍 👍 👍

PR #286, opened 2017-05-20
Macro buttons

johnbenweber:
I don't think this has been asked for but I noticed it in Universal Gcode Sender and thought it was a cool feature. The buttons can be bound to any text string through the settings menu allowing. Only about 5 minutes of coding so no big deal if you don't want it.


BarbourSmith:
Tested with hardware, looks great! 

We may re-claim that space on the front page at some point if we need the buttons, but I think the idea of having macros available is fantastic.

Issue #285, opened 2017-05-19
GC0.72 fails to render G3 properly


blurfl:
G3 gcode commands are drawn as straight lines between the endpoints. An example can be seen with this program:

G20
G1 Z0.25
G1 Y-2 F25 
G1 Z-0.1
G3 Y2 J2.0
G3 Y-2 J-2.0 
G1 Z0.25
G1 X0 Y0
G1 Z0

blurfl:
GroundControl0.71 works correctly, the issue seems to have come with version 0.72

johnbenweber:
Try PR #287.

blurfl:
PR #287 does indeed resolve this, Very well done 💯 👍

BarbourSmith:
This has to do with the way  lines are rendered. Making each line of gcode it's own object on the canvas was really slowing the program down, now just the points that make up the line are stored as part of one continuous object. This is simple for lines, but for curves it takes a little more work.

BarbourSmith:
This should be fixed in #293

blurfl:
👍

Issue #200, opened 2017-04-14
NC file too big to load due to "Follow the Path" lettering


blsteinhauer88:
Here is the file that would not load.
[Bench Back For Kids (2).zip](https://github.com/MaslowCNC/GroundControl/files/923125/Bench.Back.For.Kids.2.zip)

BarbourSmith:
I'll dig into it and see what I can find out! Thank you for sharing it.

BarbourSmith:
@blsteinhauer88 Are you sure that's the one that doesn't load? That one looks to me like it's just the outline, but it seems to load ok

![outline](https://cloud.githubusercontent.com/assets/9359447/25097013/936c9ffc-2357-11e7-8d68-7aefc5fa8284.JPG)

blsteinhauer88:
[Bench Back For Girls Final.zip](https://github.com/MaslowCNC/GroundControl/files/926406/Bench.Back.For.Girls.Final.zip)
Try This one, big!

BarbourSmith:
That looks like the one :-)

I'm seeing 50,000+ lines, and it doesn't load

BarbourSmith:
This one is going to be a little tricky. I've got figure out what's slowing everything down and it seems to have to do with the way the underlying graphics system is processing everything. 

As a note to my future self, the first thing to look into is to if rendering the whole thing as one polyline iwth 50,000+ points will render faster than 50,000 individual lines. 

I've added an error popup, and at least the file will try to load now, but on my computer GC gets VERY slow. Added in #205

BarbourSmith:
Support for big files (up to 60,000 lines) added in #279 

![big file](https://cloud.githubusercontent.com/assets/9359447/26029961/b11ec4ee-37f7-11e7-8645-50da94a184b3.JPG)

Thanks for suggesting the change, sorry it took me so long to make it. 

Good to close?

blsteinhauer88:
Yes looks good!

PR #287, opened 2017-05-20
Uncommented arc line drawing

johnbenweber:
Looks like this fixes the issue from #285. I think this got commented out when bar was working on #200. I checked with the example code given by @blurfl and the big file from @blsteinhauer88 and both seem to work now.

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -307,7 +307,7 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
             #Draw arc for G02 and G03
             with self.scatterObject.canvas:
                 Color(1, 1, 1)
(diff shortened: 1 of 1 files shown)


BarbourSmith:
@johnbenweber I just merged a new pull request #293 to address the same issue.

Putting back in the commented out line would solve the problem at the cost of using more RAM. #293 does the same thing, but by breaking up the arc into points which can be added to the existing line, it should use less RAM.

Are you OK with me closing this one? Thank you for pointing out exactly what needed to be fixed.

johnbenweber:
I figured that was probably the case. Don't let me slow you down, my fix was really just meant to be a temporary stop gap.

BarbourSmith:
And it was a great stop gap! Also having this open pull request forced me not to forget that I hadn't solved the issue

PR #291, opened 2017-05-23
Make return to center not lower bit

BarbourSmith:
I've found that I very rarely am happy that the bit has been lowered
once the machine has returned to the center.


blurfl:
I like this, I've been bitten more than once 👍

PR #293, opened 2017-05-26
Fix rendering arcs

BarbourSmith:
Fix for issue #285 same fix as #287 but with speed increase still in place


BarbourSmith:
This pull request restores rendering of G2 and G3 commands, but does it in a way which still leverages adding points to a line, rather than creating an entirely new line object which is RAM intensive.

This could be further accelerated by using fewer points on small arcs.

davidelang:
On Thu, 25 May 2017, BarbourSmith wrote:

> This could be further accelerated by using fewer points on small arcs.

For the purposes of Ground Control rendering, I would suggest that you add a 
small function to anything that add a point to a line that checks the distance 
from the last point, and only adds the point if the distance is greater than 
some minimum (possibly tuned by the zoom level and canvas size if you know this)

Even if it's a bunch of straight line segments being added, it doesn't make 
sense to plot them if they are too small to see (less than one pixel on the 
display)

BarbourSmith:
I agree. That's a good way to do it.

Issue #299, opened 2017-05-27
Prev and Next buttons in 'Actions -> About' crashes GC


GeroBH:
Both buttons cause GC to crash with "TypeError: 'NoneType' object is not callable".
Perhaps these buttons are obsolete.

BarbourSmith:
I see this too. I think these buttons were added to allow you to flip through the pages of the gcode file in view -> gcode.

Good catch.

johnbenweber:
Ah, I didn't notice there was another use of scrollabletextpopup when I did that. See PR #305.

PR #305, opened 2017-05-31
Fix for About popup

johnbenweber:
See issue #299. I just made a new class for the pageabletextpopup and reverted the scrollabletextpopup to its previous behavior.


BarbourSmith:
Looks great! Awesome work 💯 💯 👍 👍

Issue #300, opened 2017-05-28
Error while cutting complex curve.


blurfl:
I moved this here from Firmware#234 

Using GC with PR#298 and Firmware from PR#233 to cut a file that amounts to five 12MM diameter pocket cuts 3mm deep. I originally saw the error using metric coordinates so I created a new version using inch values. The error occurred with the inch version as well, though at a slightly differewnt line in the file.
Home has been moved 5" in the Y direction. The error seems to have occurred while executing line 54, the cutter headed west, clearly in error.
Here is the section of the file showing that line:
<img width="637" alt="gc view gcode see line 54" src="https://cloud.githubusercontent.com/assets/1851315/26531454/438d6b4e-439e-11e7-86a4-5b7fcd52240c.png">

Here is the control panel showing the interpretation. The last correct line looks like it was line 53, the following line is incorrect:
<img width="745" alt="gc executing gcode ending with line 54" src="https://cloud.githubusercontent.com/assets/1851315/26531457/55ad9ec0-439e-11e7-97ac-8cf7ba0e291c.png">

This is the file in use when the screen shots were made:
[Magnet array 12mm x 6cm x 5-SAE.nc.zip](https://github.com/MaslowCNC/GroundControl/files/1034425/Magnet.array.12mm.x.6cm.x.5-SAE.nc.zip)

davidelang:
there was a bug in GC that caused buffer overflow errors. This was fixed late 
yesterday, check that you have the latest version, and check to see if you have 
any errors show up while you are cutting.

blurfl:
That was PR#298, the version I'm using. First thing I did when the error happened was to check the pertinent file for the change, yes.

BarbourSmith:
It's very interesting that just the Y term disappeared completely. That could be a Ground Control issue, but it could also be a firmware thing. Let's keep the issue open here. It's so nicely documented, it would be a shame to move it.

blurfl:
On different runs I've seen the z term lost, the decimal point and part of the mantissa lost, and one time where only the letter Y was there. Was there a change to a buffer routine, or are we running low on ram?

davidelang:
there have been changes to the buffering in the last couple of weeks. There were 
errors that bar fixed in this week's release, but he may not have got all of 
them.

loosing the tail end of the command does sound like a buffer problem.

BarbourSmith:
I've actually noticed the '\n' (an invisible character that marks the end of the line) getting lost pretty regularly too.

[...]

PR #313, opened 2017-06-06
Throttle Sending to Arduino - Possible fix to #300

andymillernz:
Seeing corruption on inbound g-code commands on the Arduino, yet the serial send from Ground Control looks OK.   

Now throttling the sending in GroundControl to ensure a minimum delay (0.05 secs) between sending each g-code line as suspect Arduino serial port hardware not keeping up. 

Have left this configured at 56k baud however did also test successfully at 112k baud.


BarbourSmith:
This should be part of the discussion in #300 (I just wrote this comment so it will automatically link into that conversation)

Looks fantastic. I think @blurfl was looking for how to do exactly this.

BarbourSmith:
We should verify if this fixes the issue and merge it ahead of #314 because this looks like a better fix, and if it works we can modify #314 to run at full speed again

davidelang:
I don't think this addresses the root cause. If this was the problem, then I would expect that the grbl sender would have problems as well.

blurfl:
I just ran the GroundControl-work-out-what-s-going-on branch and saw a perfect run! It did prompt me for the z-axis at every Z code but also ran the z more correctly. Other than that, a zippy-feeling run with no errors. 😊👍

BarbourSmith:
@blurfl That's great! That branch doesn't let us take advantage of the Arduino's internal ring buffer fully (just one line is executed at a time). The changes @andymillernz is proposing here could possibly let us run at full speed without breaking things again.

I'm going to merge #314 which is the branch you just ran successfully, then explore this idea (which I think is a good one)

blurfl:
Here’s an article that discusses the inner workings of grbl including the area we’re looking at
 https://onehossshay.wordpress.com/2011/08/21/grbl-how-it-works-and-other-thoughts/

I noticed this:
"For streaming, this means the user interface should always wait for an ‘ok’ or ‘error’ response from grbl before sending a new g-code block.”

It is from 2011, tough. I don’t know whether it is up to date.

PR #314, opened 2017-06-06
Temp fix for issue #300 

BarbourSmith:
Wait until one line has finished processing before sending the next. Slow, but doesn't result in loss of data

### DIFF
--- a/Connection/serialPortThread.py
+++ b/Connection/serialPortThread.py
@@ -116,7 +116,7 @@ class SerialPortThread(MakesmithInitFuncs):
                         self._write(command)
                 
                 #Send the next line of gcode to the machine if we're running a program
-                if self.bufferSpace == 256:
+                if self.bufferSpace > len(self.data.gcode[self.data.gcodeIndex]):
                     if self.data.uploadFlag:
                         try:
                             self._write(self.data.gcode[self.data.gcodeIndex])
--- a/Connection/serialPortThread.py
+++ b/Connection/serialPortThread.py
@@ -79,7 +79,7 @@ class SerialPortThread(MakesmithInitFuncs):
             while True:
                 
                 
-                #read serial line from machine if available
+                                        #Read serial line from machine if available
                 #-------------------------------------------------------------------------------------
                 lineFromMachine = ""
                 if self.serialInstance.in_waiting > 0:
@@ -96,7 +96,11 @@ class SerialPortThread(MakesmithInitFuncs):
                     print "OK Space available: " + str(self.bufferSpace)
                 
                 
-                #Write to the machine if ready
+                
+                
+                
+                
+                                            #Write to the machine if ready
                 #-------------------------------------------------------------------------------------
                 
                 #send any emergency instructions to the machine if there are any
@@ -104,15 +108,29 @@ class SerialPortThread(MakesmithInitFuncs):
                     command = self.data.quick_queue.get_nowait() + " "
                     self._write(command)
                 
-                
+                #send regular instructions to the machine if there are any
                 if self.bufferSpace == 256:
                     print "ready"
                     if self.data.gcode_queue.empty() != True:
                         command = self.data.gcode_queue.get_nowait() + " "
                         self._write(command)
                 
+                #Send the next line of gcode to the machine if we're running a program
+                if self.bufferSpace == 256:
+                    if self.data.uploadFlag:
+                        try:
+                            self._write(self.data.gcode[self.data.gcodeIndex])
+                            self.data.gcodeIndex = self.data.gcodeIndex + 1
+                        except:
+                            self.data.uploadFlag = 0
+                            print "Gcode Ended"
                 
-                #Check for serial connection loss
+                
+                
+                
+                
+                
+                                            #Check for serial connection loss
                 #-------------------------------------------------------------------------------------
                 if time.time() - self.lastMessageTime > 2:
                     print "connection lost"
--- a/Connection/serialPortThread.py
+++ b/Connection/serialPortThread.py
@@ -86,8 +86,8 @@ class SerialPortThread(MakesmithInitFuncs):
                     lineFromMachine = self.serialInstance.readline()
                     self.lastMessageTime = time.time()
                     self.data.message_queue.put(lineFromMachine)
-                    print lineFromMachine
-                    print time.time()
+                    #print lineFromMachine
+                    #print time.time()
                 
                 #Check if a line has been completed
                 if lineFromMachine == "ok\r\n":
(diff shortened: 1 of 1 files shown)


BarbourSmith:
This is basically reverting back to how things were in 0.71. It's notably slower, but at least it gives us a stable point to work from until we find a better solution.

@blurfl Will you verify that this fixes the issue for you also?

blurfl:
Same branch?

blurfl:
GitHub service outage, will try again in a bit...

BarbourSmith:
I know, right?

Git hub being down is a big deal. They had some struggles yesterday too.

davidelang:
If you have hosting available, git repositories can be replicated, and there is 
nothing technical that makes one the master and the other the slave (just which 
one you push to first)

hopefully github will figure out what's going on here quickly, I haven't seen 
anything reported about them having problems, which surprises me a bit.

BarbourSmith:
Oops, targeted the wrong branch 😬

blurfl:
Is there a way I can help?

BarbourSmith:
I've just got to resolve some conflicts and make a new pull request.

Will you test it again once I do to make sure I haven't broken it in the process?

blurfl:
Sure! Standing by 😊👍

BarbourSmith:
Thanks! Give #315 a test when you have a chance.

I'm testing it again right now too.

blurfl:
When I run this, it prompts for the z-axis.

BarbourSmith:
Hmm, do you still see missing lines, or just the z-axis prompting?

BarbourSmith:
Prompting for the z-axis makes me think that the machine settings aren't being received by the machine right

blurfl:
No missing lines.

When I edit line 23 of serialPortThread.py as so:
        message = message + '\n'#'L' + str(len(message) + 1 + 2 + len(str(len(message))) ) + " \n"

The z axis works as expected.

BarbourSmith:
Fantastic!

What was the change to that line? I'm not spotting the difference 😀

blurfl:
I changed  “ \n”. That is double quote space backslash n double quote to single quote backslash n single quote

I know very little python, this is one of the "witch-doctor’s chicken bones" I wave at a problem. Here, it made a difference :)

BarbourSmith:
Ahhhhh! Thanks! My mind just saw quotes. I'm making the change now.

Thank you for testing that.

davidelang:
That calculation (that you are commenting out) is wrong

if the +3 causes the calculation to wrap to an additional digit, this will end 
up using more space in the queue than the L parameter says.


"(23456)" would become
"(23456)L11 \n" which is 12 bytes

BarbourSmith:
It seemed to work because we've been using it all day, but I'll keep take a look at it. Maybe something I don't understand is happening

blurfl:
Ok, I just downloaded the version with single quotes that you just did and saw the z-axis issue again.
 You mentioned machine settings, so I looked and saw the z-axis switch looked correct. I went back and poked ‘Home’ and the sled moved instead of the z axis. I went back to settings and switched z-axis off and on again, then ‘Home’ worked correctly.

davidelang:
slowing down transmissions so that the arduino can process more data makes it 
less likely to fill the circular queue if you are sending mostly 
quick-to-execute commands, so that will make it seem like it's working. But if 
you were to transmit commands that take longer to execute (large moves back and 
forth), then you will still fill up the queue and run into the problem

for example, I'll bet that something like the following will still give you 
grief (lots of commands that take a lot of time to execute)

G21 G90 G40
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000
G1 X-1140.0000000 Y-480.000000 Z0.0000000 F6.000000
G1 X1140.0000000 Y480.000000 Z0.0000000 F6.000000

blurfl:
I’ll run your torture test, but F6.000000 is too slow. I’m going to up it to F600.0000

blurfl:
I’m thinking you didn’t intend this to be tested against the version that Bar is releasing this week. Too, how about a version that doesn’t travel to the ends of the earth; that isn’t necessary to this test, but sure takes a lot of time and means the whole work area needs to be cleared :)

blurfl:
Also, GroundControl seems pretty savvy about truncating zeros, the lines come across as (after I shortened the travel and upped the speed:

G1 X-140.0 Y-80.0 Z0.0000000 F6000.000   
G1 X140.0 Y80.0 Z0.0000000 F6000.000   
G1 X-140.0 Y-80.0 Z0.0000000 F6000.000   
G1 X140.0 Y80.0 Z0.0000000 F6000.000   
G1 X-140.0 Y-80.0 Z0.0000000 F6000.000   
G1 X140.0 Y80.0 Z0.0000000 F6000.000

blurfl:
Bar, I just downloaded and ran the master after PR#315. The machine settings aren’t correctly loaded so when ‘Home’ is pressed the sled moves. A work-around is to switch the z-axis setting off then back on. The complex cut ran without error :) .

davidelang:
The purpose was to have this take lots of time, so that the firmware would not 
be clearing the buffer, even with the delays in sending.

That's why the large, slow movements.

This is intended to be run against the version that still sends multiple 
messages, not the slower-but-reliable version where only one command at a time 
is pending.
k

blurfl:
How much time should it take to fill a 256 byte buffer at 57600 baud? Traveling 2400 mm at 6mm per minute for each line of gcode does seem extreme. :)

PR #315, opened 2017-06-07
Work out what s going on

BarbourSmith:


BarbourSmith:
This is a version of #314 except targeting the right branch 😬

Issue #321, opened 2017-06-13
Z-axis calibration runs in mm when GC is set for inches, moves 1 inch when 1mm button clicked.


blurfl:
Set GC GUI to inches mode, go to Calibrate Machine Dimensions, skip to Step 7 without running any other steps. Buttons read in mm but machine has not yet been sent G21. Z-axis moves 1" or 0.1" rather than 1mm or 0.1mm. Not sure whether the issue is the button text or the mismatch between the text and the movement value.

BarbourSmith:
Fantastic catch!

Nobody wants to move 1 inch 😁 

I'll fix it right now.

BarbourSmith:
Should be fixed in #322 thanks for the catch!

blurfl:
Yup, quite a surprise 🙂

blurfl:
Fixed in #322

PR #322, opened 2017-06-13
Fix inches on calibration adjust z axis

BarbourSmith:
The calibration of the z-axis didn't switch the units to mm before moving, so if the machine was in inches it would move futher than expected


BarbourSmith:
Addresses #321

PR #333, opened 2017-06-22
Add ranchers calibration process suggestions

BarbourSmith:
-Add warning about chain wrapping to both first step of machine calibration process and chain length calibration process
-Add a 'stop' button to chain length calibration process
-Make sure that the 'home' and 'return to center' commands use the G90 command instead of the G91 command after calibration
-Add warning about chain on chain tensioner when doing the vertical measurement


BarbourSmith:
-Also, make minimizing Ground Control not close the calibration popups

PR #336, opened 2017-07-01
Avoid invalid list indexing in case no data loaded and negative index

gilbertf:
If both first two if statements apply returning because of empty list (maxIndex < 0) has to be executed.

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -124,10 +124,10 @@ class FrontPage(Screen, MakesmithInitFuncs):
         targetIndex = self.data.gcodeIndex + dist
         
         #check to see if we are still within the length of the file
-        if targetIndex < 0:             #negative index not allowed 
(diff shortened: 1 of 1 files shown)


johnbenweber:
Your description is a little confusing, I first thought maybe you didn't understand how an `if...elif` works but now I see you found a corner case, good job!

The underlying problem here appears to be that if no file is loaded and you press the skip backward button GroundControl will try to read the 0 index element of `self.data.gcode` which should be empty and likely crash.

johnbenweber:
Thinking about this some more, it might be a good idea to add `self.data.gcodeIndex = 0` to the `maxIndex < 0` case so that we ensure that we don't have some `gcodeIndex` value when no file is loaded.

BarbourSmith:
Great catch! 

Sorry it took me so long to merge this. We've been packing boxes like crazy

Issue #310, opened 2017-06-05
Truncate gcode values at 4 digits beyond the decimal point.


blurfl:
Some gcode creation programs write values to 8 or 9 digits beyond the decimal point.

 This is way beyond our precision and impacts serial transmission, buffering and string handling.
GroundControl should truncate those values and keeping four digits beyond the decimal point seems about right to me.  

Universal Gcode Sender has a configuration option to choose the number of digits but that could be an enhancement for the future.

davidelang:
another useful thing would be to calculate the distance between one location and 
the next, and if the difference is too small, skip the command (will really help 
when the source used too many line segments for a curve)

k

johnbenweber:
Wrote a fix for truncating floats to 4 digits, see PR #339.

blurfl:
Nicely done, thanks for all the effort!

PR #339, opened 2017-07-08
Added regular expression substitution to truncate long floats.

johnbenweber:
Truncate digits more than 4 from the decimal place. See Issue #310.

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -103,7 +103,11 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
             rawfilters = filterfile.read()
             filtersparsed = re.sub(r'\(([^)]*)\)','',rawfilters) #removes mach3 style gcode comments
             filtersparsed = re.sub(r';([^\n]*)\n','',filtersparsed) #removes standard ; initiated gcode comments
-            filtersparsed = re.sub(r'([+-]?\d*\.\d{1,4})(\d*)',r'\g<1>',filtersparsed) #truncates all long floats to 4 decimal places, leaves shorter floats
+
+            if self.data.config.getint('Advanced Settings','truncate'):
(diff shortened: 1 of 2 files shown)


blurfl:
Thank you, John - that’s exactly what I had in mind! Would it be useful to have this be a checkbox option, or to have the number of digits changeable, maybe on the advanced settings page? For me, 4 digits seems right but others might have different thoughts.

davidelang:
4 digits is probably right if you are in inch mode, but drastic overkill if you 
are in mm mode.
kk

blurfl:
Would we need separate settings for each mode? Would the program need to identify which system the file used, so to choose the appropriate value or would the user be responsible to maintain the corrrect setting for the file in current use? That’s all getting pretty complicated. 
I think just leaving it like this, truncating (not rounding) at four digits is the way to go.

johnbenweber:
We could add a setting for number of decimal places. I think 4 places is probably overkill for both modes. I vaguely remember @BarbourSmith saying that accuracy was ~0.5mm or ~0.020in meaning we would really be alright in both cases with 2 decimal places. I'm not sure how valuable saving characters is so it's hard to say if further complication is justified.

blurfl:
You’re right about the accuracy, two digits probably is close enough. Being able to set the digits or at least enable/disable the truncation would be nice, though.
Buffering the serial data is an ongoing issue, so saving characters might be helpful as a bandaid.

johnbenweber:
If I have time I can work on adding relevant options (enable/disable and digits).

johnbenweber:
Give it a try now. I think the problem was specifying a python boolean "True" for the default value when kivy was expecting a 1 or 0.

blurfl:
That's the bunny! Now it is perfect. 💯 👍 🥇

BarbourSmith:
Excellent addition and great community review! 

Great work everyone 👍 👍 👍

PR #341, opened 2017-07-14
Update Simulation to give us information about how errors in machine dimensions affect the final cut

BarbourSmith:


davidelang:
can this be extended to not just do a square in the center, but also in other 
places around the work area (top and bottom corners for example)

BarbourSmith:
Yes, I like that idea a lot

PR #343, opened 2017-07-14
Tweaks

davidelang:
minor tweaks to the simulation

plot the grid within the window (not centered as I don't know how to find the window width)
change the grid to 6" squares rather than the odd measurements that existed before.


davidelang:
added changes to the kv display to include a slider for CG error and display the machine dimensions.

davidelang:
I'm not sure why it is complaining about a conflict, I'm adding some lines where it is unhappy.

I completely re-worked this set of changes to split them up into different commits and clean them up

davidelang:
added 2x speedup commit

davidelang:
reading the merge commit, it looks like I had another thing that snuck in without me realizing it, tinkering with the scale size, check that if it comes out too large on your systems.

BarbourSmith:
In general these all seem like beneficial changes. I agree that we can cherry pick out the commit which moved it to the center because it's not necessary anymore, but overall awesome contributions!

PR #346, opened 2017-07-17
Make simulation load fully visible

BarbourSmith:


BarbourSmith:
Because the center of the machine is (0,0) it was loading off to the bottom left corner of the screen. I was trying to make it move to the center, but it wasn't working because I wasn't waiting for the canvas to be drawn before trying to move it. This pull request adds a delay before applying the move. It's not perfect, but I wanted to get something in place quickly so I could merge #343

davidelang:
343 also moved it onto the work area, so you probably want to think about 
skipping the first commit of #343 as that was it's only purpose. hopefully that 
won't make it too hard to merge the rest of it.

you may find it's easier to back out this change, merge 343 and then add this 
change back in.

PR #347, opened 2017-07-18
Clarification

davidelang:
This branch has two commits, the first just adds some comments and fixes it so that l=0 works (the lazy susan design)

The second commit reduces the required accuracy to be much closer to what the machine is able to do rather than being much stricter. This could cause cases where the machine tries to move to 0,0 instead of the desired location, and would also 'cannot find position, calibrate chain length' errors

We should probably change the firmware to have similar accuracy bounds.

### DIFF
--- a/Simulation/kinematics.py
+++ b/Simulation/kinematics.py
@@ -26,6 +26,12 @@ class Kinematics():
     Time = 0
     Mirror = 0
 
+    # Definition of angles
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Excellent changes 👍 👍 👍 

I agree that similar changes in the firmware side would be a good idea.

PR #374, opened 2017-07-24
Modify Simulation for Chain Slippage

krkeegan:
Simulate what it would look like if the chain position is off by a specified number of links.

This error is not symmetrical so it required running the simulation on the full work area, not just the left side and mirroring for the right.  This does result in taking twice as long to generate a simulation, but I think it is worth for the added capability.

### DIFF
--- a/Simulation/kinematics.py
+++ b/Simulation/kinematics.py
@@ -17,6 +17,8 @@ class Kinematics():
     machineHeight = 1219.2                              #this is 4 feet in mm
     machineWidth  = 2438.4                              #this is 8 feet in mm
     motorOffsetY  = 463.0                               #vertical distance from the corner of the work area to the sprocket center
+    chain1Offset  = 0                                   #number of links +,- that have slipped
+    chain2Offset  = 0                                   #number of links +,- that have slipped
 
     x = 2708.4
     y = 270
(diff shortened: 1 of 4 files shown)


davidelang:
I was thinking about this, we want to not just simulate when the chain jumps, 
but also if there is some other error (measurement nto started with the sprocket 
straight up for example, or one mount in a righ sled setup not measured right)

so please make it accept an error in mm, not just in links (Ihaven't looked at 
the code yet)
the reverse to go from x/y to chain lengths should still be able to be done only 
on one side, same with the ideal mode. It's only the red, distored view that 
needs an added calculation, and that's the forward for the distored second side.

krkeegan:
I can change it to mm rather than chains, that is easy.

You are correct, the calculations for the green and blue lines could still be symmetrical.  I can do that as well.  What is the purpose of the blue lines, I see them being drawn, but they are never apparent in my simulations?

davidelang:
the blue is what it is attempting to do, there were some differences early on 
between the blue and green lines, but that's not common any longer (I did a 1mm 
grid and had 4 spots that had problems)

blue and green will diverge when the software cannot figure out the proper 
position based on the chains (when the inverse/forward doesn't result in the 
same point)

BarbourSmith:
Is there consensus that we are ready to merge this?

krkeegan:
So there is nothing wrong with this merge.  But there are additional items and changes that we would like to make.  Given that this is just an informational tool that won't be used by many users.  I would merge it the way it is now and I will continue to work on improving my additions.

davidelang:
sounds good to me

PR #381, opened 2017-07-31
Simulation for triangular kinematics

BarbourSmith:
Add a simulation option for triangular kinematics


Simulation/kinematics.py
```diff
@@ -1,4 +1,5 @@
 import math
+from kivy.properties                         import BooleanProperty
```
BarbourSmith: not needed

Issue #383, opened 2017-08-11
No Chain Movement During Calibration


blurfl:
When kinematics.cpp sends “Message: Unable to find valid machine position. Please calibrate chain lengths.”, main.py puts up a NotificationPopup box with continueOn and hold buttons. The hold button is inappropriate, and causes user confusion by putting the machine into a state where the controls seem unresponsive.
 See discusson on [forum.](http://forums.maslowcnc.com/t/no-chain-movement-during-calibration/422/9?u=blurfl)
 Since the NotificationPopup seems only to be used for this one purpose, I think we should remove the Hold button from it to prevent the confusion.

blurfl:
This issue is solved by PR#384 👍 💯

PR #384, opened 2017-08-12
Fix issue #383

blurfl:
When kinematics.cpp sends “Message: Unable to find valid machine
position. Please calibrate chain lengths.”, main.py puts up a
NotificationPopup box with continueOn and hold buttons. The hold button
is inappropriate, and causes user confusion by putting the machine into
a state where the controls seem unresponsive.
See discusson on forum.
Since the NotificationPopup seems only to be used for this one purpose,
I think we should remove the Hold button from it to prevent the confusion.

### DIFF
--- a/UIElements/notificationPopup.py
+++ b/UIElements/notificationPopup.py
@@ -13,5 +13,4 @@ class NotificationPopup(FloatLayout):
     
     '''
     continueOn = ObjectProperty(None)
-    hold = ObjectProperty(None)
     text = StringProperty("")
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -516,9 +516,6 @@
         BoxLayout:
             #size_hint_y: 1
             Button:
-                text: "Hold"
-                on_release: root.hold()
-            Button:
                 text: "Continue"
                 on_release: root.continueOn()
 
--- a/main.py
+++ b/main.py
@@ -538,7 +538,7 @@ class GroundControlApp(App):
                     self._popup.dismiss()                                           #close any open popup
                 except:
                     pass                                                            #there wasn't a popup to close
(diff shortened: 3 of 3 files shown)


BarbourSmith:
Great job tracking down what was causing this issue!

Are we sure that we never want the "hold" option?

The other time this popup is used is when manually setting the z-axis depth. I don't know if I've ever used the hold behavior under those circumstances, but I thought I'd throw that out there.

I think we don't need it

blurfl:
I agree, we don’t need ‘Hold’ in these cases.
I wonder why I’ve been having this dialog pop up so often when starting GC lately (3 times  out of 5 - it’s getting old...)? What is Kinematics.cpp looking for that it can’t find? Is it triggering too early, before GC has sent something? Does it have to nuke my settings? Why isn’t a ‘Manual Chain Cal’ sufficient? Where should start looking?

BarbourSmith:
I've been seeing it a lot too lately. I'm not sure what's going on, but I'm going to look into it. I think it might have to do with the machine attempting to load it's position before Ground Control has sent the machine's dimensions

blurfl:
Should kinematics wait for the first B03 before calculating?

BarbourSmith:
I think it does wait for the first B03 before calculating the position, but maybe the first B03 doesn't contain all the information it needs? I started splitting the B03 command into two lines to keep it less than the 80 character GRBL line maximum. Something strange is going on for sure and I don't have a clear hunch about what it is

blurfl:
Sounds like a good spot to look into.

blurfl:
I've nmodified main.py to send the whole package of B03 settings at once and that seems to have solved the issue. More time needed to live with it, though...

blurfl:
I've done enough cutting and start/stopping GC to think that this is the root of the issue. Shall I submit a PR?

Issue #404, opened 2017-08-31
First Command Sent is Ignored


krkeegan:
I don't know the reason, but the first command sent after the Serial connection is made is never received by the Arduino.  Adding delays don't seem to help.  However sending a simple blank line seems to work but queuing this blank line doesn't always result in it being send first.  The G20 and G21 commands sent by an openfile seem to end up first, but don't seem to have any ill effects being ignored.

It is possible that the proper protocol for serial connections is to send a line return \n first.  I have seen similar things with other protocols.

If that is case, or maybe even if it isn't, we may want to consider just sending a blank line directly after the connection is made and not even queuing it to avoid the queue issue.

krkeegan:
I have never figured out why this happens, but #405 hides this from being a problem, so I am closing this.

PR #405, opened 2017-08-31
Fix Duplicate Push of Settings; Kludge to Overcome First Cmd Issue

### LINKED ISSUE #402: Kinematics posts chain re-cal required if no gcode file open
Opening a gcode file is a workaround.

[Edit] On my system it appears to matter whether the gcode file is in inches (G20) or mm (G21). If inches, the workaround is valid. If the file is in mm, the dialog appears with chain lengths 0.00 when GC is started.

A discussion on the [forum](http://forums.maslowcnc.com/t/impossible-to-cut-a-piece-without-losing-calibration/454/49?u=blurfl)

### LINKED ISSUE #403: Kinematics posts chain re-cal required if 'Settings/Advanced/Triangular' is selected
This may be caused by other settings not being adjusted, but it is not clear which ones are involved.

A discussion on the [forum](http://forums.maslowcnc.com/t/impossible-to-cut-a-piece-without-losing-calibration/454/49?u=blurfl)

krkeegan:
Bind will cause the machine data to be sent, so there is no need to force an additional call on startup.

Similarly, bind will get called on both changes in connection status, so only push setting when connection is up.

Finally, hacked in a blank line into the queue upon serial connection to deal with #404.  There may be better solutions.

Fixes #403 
Fixes #402

### DIFF
--- a/Connection/serialPortThread.py
+++ b/Connection/serialPortThread.py
@@ -90,6 +90,7 @@ class SerialPortThread(MakesmithInitFuncs):
             #print "port open?:"
             #print self.serialInstance.isOpen()
             self.lastMessageTime = time.time()
+            self.data.gcode_queue.put(' ') # send a blank line on startup
             self.data.connectionStatus = 1
             
             self._getFirmwareVersion()
--- a/main.py
+++ b/main.py
@@ -352,8 +352,6 @@ class GroundControlApp(App):
         self.data.bind(connectionStatus = self.push_settings_to_machine)
         self.data.pushSettings = self.push_settings_to_machine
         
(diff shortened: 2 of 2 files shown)


BarbourSmith:
Fantastic!

That extra pushing of settings to the machine was in there because someone was reporting that their machine wasn't getting settings pushed on connection. I fully support going back to using the bind option (because that's the right way, and if it's not working we should figure out why instead of just sending everything twice), but let's keep an eye on it.

Thanks @krkeegan!

krkeegan:
I suspect they had an issue because of #404.

PR #408, opened 2017-09-04
Ignore decimal places in gotoLine popup, enable keyboard input for NotificationPopup and TouchNumberInput popups

wrh3c:
Please review my two commits for possible inclusion in GroundControl.

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -325,7 +325,7 @@ class FrontPage(Screen, MakesmithInitFuncs):
         
         '''
         try:
-            line = int(self.popupContent.textInput.text)
+            line = int(float(self.popupContent.textInput.text))
             if line < 0:
                 self.data.gcodeIndex = 0
             elif line > len(self.data.gcode):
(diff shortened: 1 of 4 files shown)


blurfl:
This makes sense. Which platform has it been tried on? I can test OS X and/or Win10, if needed.

BarbourSmith:
@blurfl I'm about to test windows 10, if you could do OS X and give us your feedback that would be awesome!

BarbourSmith:
Checks out GREAT on windows. Thanks @wrh3c  for taking the time to make sure the callbacks for the canvas zoom don't break with the new popup feature. It feels really natural to me to type in the values like this, and I really like that the enter key closes the popup and applies the value.

Excellent contribution. Thank you! 👍 👍 ❤️ ❤️

blurfl:
Will do. I’m on a ladder, get it in 45 minutes.

BarbourSmith:
@blurfl  be safe up there, we need you 😁

blurfl:
Works fine on OS X, a very good improvement. 💯 
 'on a ladder'... I should have said "I've gotta climb back up and finish before sitting down to a keyboard."

PR #411, opened 2017-09-06
Enable keyboard input for z-axis popup

wrh3c:
Following up to my previous pull request, here is an addition to enable keyboard input for the z-axis popup for numeric entry.

PR #412, opened 2017-09-06
Fix popup numeric display

wrh3c:
Fix the number displayed after closing a numeric input popup.
Currently, the number displayed is the text version (string) of what was entered.
For the fix, convert the entered sting to a float (which was already being done to determine whether the entry was a valid number), then convert the float back to a string to be displayed.

Examples:
User Entry, Before Fix, After Fix
"000",        "000",       "0.0"
".1",           ".1",           "0.1"
"1.000",     "1.000",     "1.0"


blurfl:
A nice refinement, thanks for separating these. Tested OK on OS X

wrh3c:
I tried merging this along with #411 into a "clean" copy of master in my repository, and had some merge issues which I believe are due to the reverted commits above.  I'll be more careful with keeping changes separated in branches going forward.  If GitHub wants to change (delete and add) more than 2 lines in each of 2 files for this pull request, it might be better to make the changes manually.

blurfl:
GitHub can be something of a mystery to me too. Thanks for taking the trouble to unwind this PR, and for the excellent contributions you're making.

wrh3c:
Okay, I made "clean" versions of pull requests #411 and #412, with just one commit in each, which merged smoothly into a copy of the latest master.  Should I close these two pull requests and submit the clean versions?

blurfl:
Yes, if you would. Thanks very much.

wrh3c:
Okay, thank you for your patience and words of encouragement, and sorry for the churn.  I opened pull request #414 with a single clean commit against the original master, and I'm closing this one.

blurfl:
Churn all you like, if you bring such improvements 💯 👍

PR #415, opened 2017-09-08
Update Hold button and stop sending Gcode on pause from Maslow

wrh3c:
When Maslow sends "Maslow Paused", set uploadFlag to 0.  This updates the "Hold" button to say "Continue" and stops sending Gcode commands.

While this might not be strictly necessary right now since the "Message:" handler also sets uploadFlag to 0, it could be helpful in the future if/when there are other things that might pause Maslow.  For example, I'm working on Tool Change handling which pauses Maslow while the user changes the tool.  If I use a version of Ground Control that doesn't include a pop-up to prompt the user (but does include this change), the button saying "Continue" instead of "Hold" is a visual indication that Maslow is paused, so the user can look to the text console to see why.  Also, when the user wants to continue, clicking the "Continue" button (once) will un-pause Maslow.  If the button isn't updated, the user has to click it twice: once to change it from "Hold" to "Continue", and again to have Ground Control send the command to un-pause Maslow.

### DIFF
--- a/main.py
+++ b/main.py
@@ -535,6 +535,9 @@ class GroundControlApp(App):
                     print message
                     measuredDist = float(message[9:len(message)-3])
                     self.data.measureRequest(measuredDist)
+            elif message[0:13] == "Maslow Paused":
+                self.data.uploadFlag = 0
+                self.writeToTextConsole(message)
             elif message[0:8] == "Message:":
                 self.previousUploadStatus = self.data.uploadFlag 
                 self.data.uploadFlag = 0


wrh3c:
I'm also working on implementing some M codes, the first two of which are program pauses.  Rather than trigger a popup notification, I would like to have Ground Control just pause with the "Hold" button changed to "Continue" and wait for the user to click "Continue".

BarbourSmith:
Fantastic! We actually already ran into this issue already because it was possible to pause the machine from within the popup for adjusting the z-axis, but it wasn't clear how to unpause it. We just ended up taking out the button, but your solution would have fixed it 👍 👍 👍

main.py
```diff
@@ -535,6 +535,9 @@ def runPeriodically(self, *args):
                     print message
                     measuredDist = float(message[9:len(message)-3])
                     self.data.measureRequest(measuredDist)
+            elif message[0:13] == "Maslow Paused":
```
BarbourSmith: Is there a cleaner way to do this than looking at a range of characters? I know this is my fault (I started it), but I feel like maybe we should be doing a search to see if the string "Maslow paused" appears in the command or something

wrh3c: I was just following the example from the other parts of the if statement.  These will be the first characters in the message, so I think we're okay.  If you want to experiment with a string search, that's fine too!

PR #416, opened 2017-09-08
Write Ground Control version to log file

wrh3c:
Okay, let's see if I can keep this one simple!  This should have only one commit deleting one line and adding two lines to main.py.  (The only reason the one line shows as a deletion is because I changed its indentation level by one space to line up with all the other lines at this level.)

When the Firmware version is reported, Ground Control adds its version and displays both on the text console.  This change also writes the Ground Control version to the log file.  (The Firmware version is already written to the log file before the if statement.)


blurfl:
Makes good sense, 👍

BarbourSmith:
Looks fantastic! Great idea

Issue #400, opened 2017-08-25
Escape button causes GroundControl to quit.


JDogHerman:
clicking the escape button...

```[INFO   ] [Base        ] Leaving application in progress...
 Traceback (most recent call last):
   File "main.py", line 643, in <module>
     GroundControlApp().run()
   File "C:\Python27\lib\site-packages\kivy\app.py", line 828, in run
     runTouchApp()
   File "C:\Python27\lib\site-packages\kivy\base.py", line 504, in runTouchApp
     EventLoop.window.mainloop()
   File "C:\Python27\lib\site-packages\kivy\core\window\window_sdl2.py", line 663, in mainloop
     self._mainloop()
   File "C:\Python27\lib\site-packages\kivy\core\window\window_sdl2.py", line 606, in _mainloop
     self.modifiers)
   File "kivy\_event.pyx", line 714, in kivy._event.EventDispatcher.dispatch (kivy\_event.c:8146)
   File "kivy\_event.pyx", line 1225, in kivy._event.EventObservers.dispatch (kivy\_event.c:14035)
   File "kivy\_event.pyx", line 1149, in kivy._event.EventObservers._dispatch (kivy\_event.c:13564)
   File "C:\Python27\lib\site-packages\kivy\app.py", line 1040, in _on_keyboard_settings
     return self.close_settings()
 TypeError: close_settings() takes exactly 2 arguments (1 given)

wrh3c:
I believe this has been fixed by pull request #408.  Now that the keyboard binding in GcodeCanvas is preserved even when popups are opened and closed, the Esc key is caught and ignored.

That said, we should consider setting the kivy exit_on_escape config token to 0 to prevent the Esc key from quitting GroundControl.  See the discussion at https://stackoverflow.com/questions/27590127/leaving-kivy-app-when-pressing-escape for more information.

BarbourSmith:
I think we might have wanted it to quit the program for those operating on raspberry pi where Ground Control launches full screen with no option to click an "X" in the corner.

Does anyone else remember that? @blurfl ? @TheRiflesSpiral ?

blurfl:
I don’t remember, and I’m not in a place to check. There is the quit button though.

BarbourSmith:
Great memory, I forgot that we also added the quit button. That should be enough.

TheRiflesSpiral:
Yep, this is how I remembered it. Esc to navigate back through menus, Quit button to quit the app.

[...]

PR #417, opened 2017-09-09
Enable Esc key and fix issue 400

wrh3c:
Enable use of Esc key to navigate back through menus by letting unhandled keys in gcodeCanvas.py be passed to the next callback.  This behavior was broken in pull request #408.  This change also allows use of the F1 key to enter and exit the settings menu.

Now the Esc key causes Ground Control to crash, per issue #400, when pressed at the main screen or in the settings menu.  The same crash is caused by the F1 key when exiting the settings menu.  The issue is caused by the close_settings() handler in main.py.  The error is "TypeError: close_settings() takes exactly 2 arguments (1 given)"  Since close_settings() in main.py doesn't do anything besides call the super version of itself, deleting it solves the problem with no loss of functionality.

Finally, the Esc key will quit Ground Control if pressed at the main screen.  We solve this by setting the kivy exit_on_escape config token to 0 in main.py.

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -85,10 +85,13 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
         if keycode[1] == self.data.config.get('Ground Control Settings', 'zoomIn'):
             mat = Matrix().scale(1-scaleFactor, 1-scaleFactor, 1)
             self.scatterInstance.apply_transform(mat, anchor)
+            return True # we handled this key - don't pass to other callbacks
         elif keycode[1] == self.data.config.get('Ground Control Settings', 'zoomOut'):
             mat = Matrix().scale(1+scaleFactor, 1+scaleFactor, 1)
             self.scatterInstance.apply_transform(mat, anchor)
-        return True
+            return True # we handled this key - don't pass to other callbacks
+        else:
+            return False # we didn't handle this key - let next callback handle it
 
     def isClose(self, a, b):
         return abs(a-b) <= self.data.tolerance
--- a/main.py
+++ b/main.py
@@ -5,6 +5,7 @@ Kivy Imports
 '''
 from kivy.config                import Config
 Config.set('input', 'mouse', 'mouse,disable_multitouch')
+Config.set('kivy', 'exit_on_escape', '0')
 from kivy.app                   import App
 from kivy.uix.gridlayout        import GridLayout
 from kivy.uix.floatlayout       import FloatLayout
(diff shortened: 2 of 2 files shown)


wrh3c:
I noticed this behavior which initially concerned me, but I believe it is desirable:

1. Click "Z-Axis" to open the Z-Axis menu (6 buttons)
2. Click the top left button ("0.1" by default) to open the numeric entry popup
3. Enter a new number and click "Done"
4. Note that the new number is displayed in the top left button
5. Press the Esc key to exit the Z-Axis menu
6. Click "Z-Axis" to open the Z-Axis menu again.  Note that the number displayed in the top left button has reverted back to the original value.

This behavior is different than clicking the "Done" button to exit the Z-Axis menu.  Replace steps 5 and 6 above with:

5. Click the "Done" button to exit the Z-Axis menu
6. Click "Z-Axis" to open the Z-Axis menu again.  Note that the number displayed in the top left button is the new one that was entered in step 3 and displayed in step 4.

I believe this is consistent with Esc meaning "escape", i.e. "get me out of here without changing anything!"  If the user presses Esc in the numeric entry popup, any number entered is ignored and the original value is preserved.  However, the user did click "Done" to accept the new number and exit the numeric entry popup, so should this new number be preserved even if the user presses Esc to exit the Z-Axis menu?

blurfl:
I agree that the escape “get me out of here without changing anything!" behavior is desired. 👍

wrh3c:
The only concern I have is this behavior is different from using Esc to close a menu.  If I go into the settings menu, change a setting, then press the Esc key to exit the menu, the setting I changed is preserved.

On the contrary, pressing the Esc key to exit the Z-Axis popup does not preserve the change I made to the numeric value.  Obviously, it won't undo any movement commands initiated from the Z-Axis popup, so I believe it is only the numeric value that is affected.

If everyone is okay with this behavior, we should be good to go!

UIElements/gcodeCanvas.py
```diff
@@ -85,10 +85,13 @@ def _on_keyboard_down(self, keyboard, keycode, text, modifiers):
...
             mat = Matrix().scale(1-scaleFactor, 1-scaleFactor, 1)
             self.scatterInstance.apply_transform(mat, anchor)
+            return True # we handled this key - don't pass to other callbacks
         elif keycode[1] == self.data.config.get('Ground Control Settings', 'zoomOut'):
             mat = Matrix().scale(1+scaleFactor, 1+scaleFactor, 1)
             self.scatterInstance.apply_transform(mat, anchor)
-        return True
+            return True # we handled this key - don't pass to other callbacks
+        else:
+            return False # we didn't handle this key - let next callback handle it
```
BarbourSmith: This works out really elegantly! Nice!

BarbourSmith:
I'm OK with the distinction, but I really really appreciate your taking the time to think deeply about what the right behavior is 👍 👍 👍 👍

PR #418, opened 2017-09-09
Allow windowed mode

wrh3c:
Under the settings for kivy, there is an option for Fullscreen, which allows the user to choose windowed or fullscreen mode.  But Ground Control forces fullscreen mode regardless of this setting by calling window.maximize() in the build() function of main.py.  This change removes window.maximize() to allow the user to choose windowed or fullscreen mode.

A side effect of this change is that the Gcode canvas is not centered in the window.  This is because the call to window.maximize() triggered the on_resize() callback.  Now that the window.maximize() call is removed, the callback does not occur at initialization.  To fix this, a call to self.centerCanvas() is added to the initialize() function of gcodeCanvas.py.

### DIFF
--- a/UIElements/diagnosticsMenu.py
+++ b/UIElements/diagnosticsMenu.py
@@ -29,7 +29,7 @@ class Diagnostics(FloatLayout, MakesmithInitFuncs):
                     'along with the Maslow Control Software. If not, see <http://www.gnu.org/licenses/>.'
                 
         content = ScrollableTextPopup(cancel = self.dismiss_popup, text = popupText, markup = True)
-        self._popup = Popup(title="About GroundControl", content=content, size_hint=(0.5, 0.5))
+        self._popup = Popup(title="About GroundControl", content=content, size=(520,400), size_hint=(None, None))
         self._popup.open()
     
     def dismiss_popup(self):
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -44,7 +44,7 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
         
         self.drawWorkspace()
 
-        if self.data.config.getint('Ground Control Settings', 'centerCanvasOnResize'):
+        if self.data.config.getboolean('Ground Control Settings', 'centerCanvasOnResize'):
             Window.bind(on_resize = self.centerCanvas)
 
         self.data.bind(gcode = self.updateGcode)
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -43,18 +43,18 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
         self.targetIndicator.color = (1,0,0)
         
         self.drawWorkspace()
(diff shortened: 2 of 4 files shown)


wrh3c:
At smaller window sizes, some of the user interface elements which do not resize (the controls at the upper right and the text console beneath the controls) take over a large portion of the window and may be cut off by the bottom of the window.  This is also a problem when using Ground Control in fullscreen mode on a lower-resolution screen, which some people may want to do, i.e. use an old netbook or tablet to control Maslow.  By running in windowed mode, it is easy to change the window size to test the appearance of the user interface elements for different window sizes / screen resolutions.

wrh3c:
Added a second commit with additional changes for working in windowed mode:
- Changed menu and button heights to 60 pixels (groundcontrol.kv)
- Set minimum window width and height to just under 640x480 pixels (main.py)
- Added a setting under Ground Control Settings so user can choose whether to center the Gcode canvas when the window is resized (main.py and gcodeCanvas.py)
- Changed notification popup to a fixed pixel size because it was too small at 35% of a 640x480 window (main.py)
- Changed About popup to a fixed pixel size because it was too small at 50% of a 640x480 window (diagnosticsMenu.py)
- Added centerCanvasAndReloadGcode() function and bound to gcodeFile so the canvas will be zoomed out and recentered when a new Gcode file is loaded (gcodeCanvas.py)

wrh3c:
And a third commit - I just learned about the config.getboolean() function.  This change uses config.getboolean() instead of config.getint() to get the boolean value for centerCanvasOnResize.

wrh3c:
So this isn't as useful as I originally thought.  I found out that kivy requires OpenGL 2.0 or better, and my old netbook's built-in graphics card/driver only supports OpenGL 1.x.

I do like being able to choose windowed or fullscreen mode.  Even on bigger screens, it's helpful to have Ground Control in a window so I can easily see and use other windows.  But if you don't like the other changes (menu, button, popup sizes) let me know and I will close this pull request and submit a simpler one with just the changes to enable user selection of windowed vs. fullscreen mode.

BarbourSmith:
I still think it's a cool idea and that exploring making GC render right one lower resolution screens is valuable. I'd like to be able to run it on Android soon and that would need smaller screen support.

I'm not totally sure I understood the original issue. For me Ground Control always launched "full screen", but it could be made smaller by clicking the button next to the "X"

image

Did you not have that?

wrh3c:
Yes, I didn't notice that before, but I do have the zoom button!  So I guess that explains the original issue - I didn't notice a difference when launching Ground Control in windowed vs. fullscreen mode because it always took over the whole screen.

BarbourSmith:
So do we have a final verdict on this one? My vote is for merging it. I really like the bigger buttons for actions and settings, and I think that starting to think about how things will resize/show on small screens is a good thing to be doing.

Any opposition?

krkeegan:
I note that things also resize oddly on very large screens, the simulation page for one is an issue.

BarbourSmith:
Yeah, the calibration popup too. We need to spend more time on all of that.

PR #433, opened 2017-10-05
Renames the z-in and z-out buttons in the popup

### LINKED ISSUE #432: Z-Axis direction labels
This is a cosmetic/consistency thing, but it would be nice if the Z-axis controls said the same thing in the different places they appear.

In the machine calibration/setup screen they say "raise" and "lower" (which I like). On the home screen when you hit the z-axis popup control box, you get "in" and "out" which I don't find intuitive without thinking about it.

Anyhow, thought I would mention it.

designfrontier:
Fixes #432

This renames z-in to lower and z-out to raise
to improve readability of the interface on the
zaxis popup from the front page.

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -14,8 +14,8 @@
         pos: (0, root.height - 60)
         size_hint: (.5,None)
         height: dp(60)
-        
-        Button: 
+
+        Button:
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Beautiful! Thank you!

PR #434, opened 2017-10-16
Gobble space runs

blurfl:
Some gcode creation programs sprinkle space runs within the gcode lines. This will condense all runs to a single space.


BarbourSmith:
Elegant solution, and great job figuring out what the issue was. I spent the better part of a day trying to find out what was causing the issue and got nowhere.  It was still on my to-do list so thank you for clearing it up before I even got to it. 😀

blurfl:
Gero gets the credit here, he had the patience to find the cause. With such a pinpoint indicator, the rest was simple.

BarbourSmith:
Beautiful team work!

PR #437, opened 2017-10-25
Fix osx notification size

blurfl:
OSX screen resolution makes popups h- and v-size too to small to read content. Use size_hint to scale popup to the screen size, special case for darwin to avoid affecting other platforms.


BarbourSmith:
❤️ ❤️ ❤️ 

Excellent! 

This feels like a mistake on my part to have used pixel specific numbers in the first place. Is there any reason we shouldn't use size hint on every platform?

blurfl:
Popup doesn't understand the pixel specific (dp, sp) sizes, so this is the only universal way to control the size. Not sure about the other platforms, so I special-cased it. Getting rid of the special case would be cleaner 😉

BarbourSmith:
Got it. 

I agree, let's use the new method for all OS's. I think that's how Kivy would have wanted us to do it 😀

PR #438, opened 2017-10-25
Fix remove comments

blurfl:
Preserve the newline when removing comments to prevent moving multiple gcode lines into one line. Combining lines could change separate movements into a coordinated movement.
 Remove blank \n\n lines introduced when removing lines containing only a comment.


BarbourSmith:
👍 👍 👍  Excellent! I had seen this happen, but I didn't know why it happened. Great catch 🥇

PR #442, opened 2017-10-31
Move sled to center of work area after finishing measure out chains step

BarbourSmith:
Send a B15 command to move the sled to the center after the chain length step of the calibration process. A part of firmware issue [#323](https://github.com/MaslowCNC/Firmware/issues/323)


blurfl:
Does this replace the step which moved the chains to 1650mm? That would mean re-painting the 'king link' every time the sled is changed, like going from the quad- to the triangular- or from the upper-pantograph- to the ring-sled, or it would mean running the motors every time the chain cal needs to be re-established
😔

BarbourSmith:
Great question!

No, it does not replace that step or interfere with the 'manual calibration' system. Or at least I hope it does not.

The way the new system works is that the chains are fed out to 1650mm and stop the same as before. The difference is that now when the 'Finish' button is pressed at the end of the step the chain lengths are adjusted to bring the sled to the center. 

That was the best compromise I could come up with to try to meet all the goal, but I'm open to suggestions 👍

blurfl:
That's a very good compromise. 👍 👍 
 It might be an idea to mention marking the top links (I really should load the code and see for myself, sorry for being lazy here) as an option at that point.

PR #457, opened 2017-11-18
Issue 453 - Don't crash on viewing empty Gcode file

### LINKED ISSUE #453: Actions -> View Gcode without any gcode Crashes
There needs to be a check if there is any Gcode is loaded when clicking this button.
The application crashes.  It should just show an alert that there is no gcode or a blank screen.

skiz:
Resolves issue of null pointer in title which fixes #453


BarbourSmith:
Fantastic! Thank you! 👍 👍 👍 👍

PR #458, opened 2017-11-18
Added zAxisAuto configuration option for spindle automation

skiz:
Add an option to the Settings panel which enables automated power control (M codes) to be handled by the AUX1 high signal.  Accompanying firmware change is at https://github.com/MaslowCNC/Firmware/pull/328

### DIFF
--- a/main.py
+++ b/main.py
@@ -119,6 +119,13 @@ class GroundControlApp(App):
             "key": "zAxis"
         },
         {
+            "type": "bool",
+            "title": "Spindle Automation",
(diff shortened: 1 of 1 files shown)


BarbourSmith:
How do we feel about putting this in the Advanced Settings? I feel like the less options we can give everyone up front the better (not that there aren't a bunch of other options in the settings which need to be cleaned up or moved to advanced settings)

skiz:
Easy enough to change. Will do.

Issue #454, opened 2017-11-17
Calibrate Chain Length Automatic - No way to exit


jbnimble:
After opening the dialog, there is no way to exit the dialog without exiting Ground Control or Set Zero button, if decide not to perform calibration it would be convenient to not have to close Ground Control to exit the dialog.

skiz:
#459 closes #454

PR #459, opened 2017-11-18
Add a close button to chain calibration

skiz:
Add close button to calibration first step.

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -1133,6 +1133,9 @@
                 Button:
                     text: 'Set Zero'
                     on_release: root.setZero()
+                Button:
+                    text: 'Close'
+                    on_release: root.done()
                 Label:
         GridLayout:
             cols: 2
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -739,6 +739,9 @@
                 Button:
                     text: "Begin"
                     on_release: root.defineInitialState()
+                Button:
+                    text: "Close"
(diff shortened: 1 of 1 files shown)


skiz:
fixes #454

blurfl:
This would be by way of cancelling the calibration, right? Would 'Cancel' be more appropriate than 'Close'?

skiz:
@blurfl The only 'Cancel' in the app is in the save dialog, all popups use 'Close', so just following the existing design.

blurfl:
Some of these dialogs put the machine into relative mode, let's make sure that exiting this way restores the machine to the condition it was in before it entered the page.

BarbourSmith:
I like it!

@blurfl is right that we have to be careful of the machine's state when exiting the calibration routine at weird points because once the "Begin" button is pressed things like relative mode, mm mode, and the chain lengths can be left in strange states. I feel good about the close button here in the first step because it is accessible before clicking the begin button. Using the back button to get back there then closing could be an issue, but that's already possible when fast forwarding to the end

blurfl:
"...Using the back button to get back there then closing could be an issue, but that's already possible when fast forwarding to the end" This is probably how I get bitten, trying to use parts of the process. Do we need an issue to capture this?

BarbourSmith:
Yes! Great suggestion @blurfl 

I will merge this as is and let's make a new issue for "it's not clear that exiting the calibration process part way through can cause trouble"

PR #462, opened 2017-11-18
Support touchscreen number entry for calibration. Fixes #443

### LINKED ISSUE #443: Make calibration process touch screen compatable
The calibration process is the only step with requires a keyboard to enter numbers for the distances. It would be great to be able to run Ground Control entirely without a keyboard

skiz:
This adds the number popup for calibration use, and text clarifies usage.

I think we will want to move this popup to it's own element as it's beginning to be reused in a few places now.


BarbourSmith:
Looks fantastic! Thank you for making my massive ToDo list shorter! ❤️ ❤️

PR #471, opened 2017-11-25
Fix for 470

davekopecek:
Fixed typo `Aixs` = `Axis`


BarbourSmith:
Thanks for the fix, I'm embarrassed that slipped through for so long!...and WELCOME!!

PR #474, opened 2017-11-28
Fix issue#473 add spaces between gcodes

blurfl:
Operates on gcodes [GXYZIJFTM] - did I miss any?


BarbourSmith:
So quick! Someone finds and issue and you are on it in a matter of hours! Great work!! 👍 👍

BarbourSmith:
I think GXYZIJFTM covers them all

blurfl:
Not touched in this issue were
 - the question of how GC treats repeated lines of Xnn Ynn without a G1 a the start
 - in working with this file, I noticed that the renderer hiccups on a line with only an X or Y - it uses a zero for the missing operand when it draws the cursor.

davidelang:
while lines like this are legal g-code, lots of things are going to have 
problems with them. If there is a post-processor option that always includes g1 
at the start of every movement, that should be used and we should blacklist the 
one that skips it.

blurfl:
CNC_Functions.h executeGcodeLine() starts with the comment:
"Executes a single line of gcode beginning with the character 'G'.  If the G code is not included on the front of the line, the code from the previous line will be added."
 Looks like it's handled. 👍

Issue #475, opened 2017-11-28
Lines missing X or Y don't render correctly in single-step preview.


blurfl:
Presently, if X or Y are absent we use self.gcodecanvas.positionIndicator.pos[] values, but these are over-written by the real-timeposition of the router. This causes the cursor to jump off the cutline on the screen when stepping through the line missing the X or Y value.
 If a gcode line does not contain either an X or a Y, we need the previous value to correctly render the cursor during the single-step preview.

PR #476, opened 2017-11-28
preserve X and Y positions - issue #475

blurfl:
Presently, if X or Y are absent we use self.gcodecanvas.positionIndicator.pos[] values, but these are over-written by the real-timeposition of the router. This causes the cursor to jump off the cutline on the screen when stepping through the line missing the X or Y value.
If a gcode line does not contain either an X or a Y, we need the previous value to correctly render the cursor during the single-step preview.

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -31,6 +31,9 @@ class FrontPage(Screen, MakesmithInitFuncs):
     
     numericalPosX  = 0.0
     numericalPosY  = 0.0
+
+    previousPosX = 0.0
+    previousPosY = 0.0    
     
     stepsizeval  = 0
     zStepSizeVal = .1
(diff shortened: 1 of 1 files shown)


UIElements/frontPage.py
```diff
@@ -160,20 +163,16 @@ def moveGcodeIndex(self, dist):
             x = re.search("X(?=.)([+-]?([0-9]*)(\.([0-9]+))?)", gCodeLine)
             if x:
                 xTarget = float(x.groups()[0])
+                self.previousPosX = xTarget
             else:
-                if self.data.units == "INCHES":
-                    xTarget = self.gcodecanvas.positionIndicator.pos[0] / 25.4
-                else:
-                    xTarget = self.gcodecanvas.positionIndicator.pos[0]              
+            	xTarget = self.previousPosX
```
BarbourSmith: 👍 elegantly done!

BarbourSmith:
Fantastic!! I like the way you did this

Issue #460, opened 2017-11-18
Actions -> Calibrate Chains Length - Automatic does not finish at 0,0


blurfl:
It looks like though logic to move to 0,0 at the end of chain calibration in the full calibration sequence, it wasn't added to the end of the the automatic calibration on the Actions page. See discussion [here](https://forums.maslowcnc.com/t/i-put-my-linkage-kit-together/854/42?u=blurfl).
Possibly add a 'Finish' button which would call root.finishChainCalibration()?

BarbourSmith:
I think this is part of a larger issue that we have the same code implemented twice. Really there should be a single instance that is referenced in both places, right? That way a change to one will show up in both places

blurfl:
You're right, that's always the best. Got to be able to use the Automatic Chain calibration without entering the full calibration sequence, though.

BarbourSmith:
Fixed in #477 👍

Issue #456, opened 2017-11-17
Perform Test Cut Anytime (or after automatic chain calibration)


jbnimble:
It would be convenient to be able to optionally perform the test cut procedure after the automatic chain length calibration. Instead of having to go through the whole 10 step calibration process to get to the point of being able to perform the test cut.

Even better it would be convenient to be able to perform the test cut at any time.

BarbourSmith:
That is a great idea. Let me think on how to best impliment that

BarbourSmith:
Great recommendation @skiz I will add a test gcode file to the repo

krkeegan:
I have been able to skip through the calibration steps to the test cut and redo it that way.  It seems to work just fine as long as you use the skip button nothing gets erased or changed in the other steps.

BarbourSmith:
I just broke the test cut out into it's own object so it should be pretty easy now to put it in a popup up under something like "Actions -> Perform Triangular Test Cut" or something similar so we can run it as a stand alone.

I agree with @krkeegan that using only the 'skip' button should work too

Issue #448, opened 2017-11-09
Auto calibration - Step 7 suggestion


bdillahu:
On step 7, where it pays out first the left chain, then the right preparatory to connect to the sled, suggest a few seconds between sides, or a "next prompt" of some kind to make sure you can get from one side to the other. I'm holding one chain to keep it from tangling, and now that other one's moving and I've got to get this one laid somewhere and get over there :-)

blurfl:
I second the motion. I dislike having to stand by the right motor waiting. Suggest a 'next' click followed by a 5 second pause to get from the mouse to the right motor 🙂

BarbourSmith:
I'm on the other side. I would never encourage anyone else to, but I usually let my calibration process run out both chains without me touching anything. I've found that as long as the first link isn't too well planted on the gear they won't wrap (but I would never encourage anyone to do the same).

What about a solution that gives us both options:

We could have two buttons one for "measure out left chain" and one for "measure out right chain". Clicking either one would start a 5 second count down followed by the chain extending. If you want the current behavior where both chains extend automatically you could just click one, then the other and it would queue up the second one to start right after the first finished.

How do we feel about that as a solution? I think I can implement that pretty easily.

bdillahu:
Works for me... 

I know in my layout it isn't the chain getting tangled at the sprocket, but rather farther down as it falls it gets caught on itself headed back up... Probably varies a good bit depending on people's frames and the angles involved.

blurfl:
"farther down as it falls" I've had that too. I've been thinking of a piece of 'stretchy string' with a hook or magnet on the end to manage the chain while it feeds in the various calibration steps.

bdillahu:
That’s a good idea...

PR #477, opened 2017-11-28
Create objects for some calibration functionality

BarbourSmith:
The behavior of rotating the sprockets to the vertical position and extending the chains to a known length are both behaviors repeated in the chain length calibration and the machine measuring processes so maintaining the same behavior in two places was causing issues. This pull request creates widgets for each behavior which can be used in both places.

It is not ready to merge, just creating the PR to be able to view the diffs more easily for now

### DIFF
--- a/UIElements/measureMachinePopup.py
+++ b/UIElements/measureMachinePopup.py
@@ -22,8 +22,6 @@ class MeasureMachinePopup(GridLayout):
         
         '''
         
-        print "THIS BIT RAN"
-        
         self.data = data
         
         self.setSprocketsVertical.data      =  data
--- a/UIElements/measureMachinePopup.py
+++ b/UIElements/measureMachinePopup.py
@@ -31,6 +31,7 @@ class MeasureMachinePopup(GridLayout):
         
         self.measureOutChains.data          =  data
         self.measureOutChains.carousel      =  self.carousel
+        self.measureOutChains.text          =  "Now we are going to measure out the chains and reattach the sled\n\nHook the first link of the right chain on the vertical tooth of the right sprocket\n as shown in the picture below\n\nThe left chain does not need to be moved, it can be left partly extended\n\nThe correct length of first the left and then the right chain will be measured out\n\nOnce both chains are finished attach the sled, then press Next\nPressing Next will move the sled to the center of the sheet.\n\nBe sure to keep an eye on the chains during this process to ensure that they do not become tangled\naround the sprocket. The motors are very powerful and the machine can damage itself this way"
     
     def backBtn(self, *args):
         '''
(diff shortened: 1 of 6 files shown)


BarbourSmith:
This is the first step towards addressing #448

BarbourSmith:
And #460 and #456

BarbourSmith:
There is an issue with this. The orient the sprockets to 12 o'clock works great because it is identical. The extend chain's is less good because the situations are different. In one case the chains need to be attached, in the other the left chain is already attached so the pictures and directions need to be updated

Here's what it looks like now:

image

I'm going to sleep on it and see what I think in the morning

BarbourSmith:
This pull request fixes #460

UIElements/measureMachinePopup.py
```diff
@@ -15,6 +15,24 @@ class MeasureMachinePopup(GridLayout):
...
     numberOfTimesTestCutRun = -2
     
+    def establishDataConnection(self, data):
+        '''
+        
+        Sets up the data connection between this popup and the shared data object
+        
+        '''
+        
+        print "THIS BIT RAN"
```
BarbourSmith: 💥

UIElements/measureOutChains.py
```diff
@@ -0,0 +1,24 @@
...
+    data              =  ObjectProperty(None) #set externally
+    text              =  StringProperty("")
+    
+    def stop(self):
+        self.data.quick_queue.put("!") 
+        with self.data.gcode_queue.mutex:
+            self.data.gcode_queue.queue.clear()
+    
+    def calibrateChainLengths(self, direction):
+        print direction
```
BarbourSmith: 💥

UIElements/setSprocketsVertical.py
```diff
@@ -0,0 +1,38 @@
...
+        self.data.gcode_queue.put("G90 ")
+    
+    def LeftCCW(self):
+        
+        self.data.gcode_queue.put("G91 ")
+        self.data.gcode_queue.put("B09 L-.5 ")
+        self.data.gcode_queue.put("G90 ")
+        
+    def RightCW(self):
+        #print "right CW"
```
BarbourSmith: 😑

UIElements/measureOutChains.py
```diff
@@ -0,0 +1,24 @@
+from kivy.uix.widget                      import   Widget
+from kivy.properties                      import   ObjectProperty, StringProperty
+
+class MeasureOutChains(Widget):
+    '''
+    
+    Provides a standard interface for making both sprockets point vertically
```
BarbourSmith: This text is wrong

PR #479, opened 2017-11-29
Add one button for extending each chain

### LINKED ISSUE #448: Auto calibration - Step 7 suggestion

BarbourSmith:
Adds one button for extending each chain to the calibration process.

Fixes #448

### DIFF
--- a/UIElements/measureOutChains.py
+++ b/UIElements/measureOutChains.py
@@ -1,5 +1,6 @@
 from kivy.uix.widget                      import   Widget
 from kivy.properties                      import   ObjectProperty, StringProperty
+from kivy.clock                           import   Clock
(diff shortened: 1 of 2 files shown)


krkeegan:
I used this the other day.  This is a really nice improvement for new users in particular.

BarbourSmith:
The community suggestions are just a non-stop source of great tweaks to make things better

PR #483, opened 2017-11-30
Use the current value of rotationRadius if it is present

blurfl:
Trying to do the calibration step with a top-pantograph linkage (129mm rotationradius), I find that choosing the Triangular calibration writes 260 to rotationradius in groundcontrol.ini overwriting the previous value. This PR looks for the current value and uses that if it is found.


blurfl:
Operator error, sorry.

Issue #337, opened 2017-07-04
Add version as top level variable in main.py for Android APK building


jbnimble:
When building an Android APK out of GroundControl it is required to have a "version" variable at the top level in the main.py. See the example in the [building executables](https://github.com/MaslowCNC/GroundControl/wiki/Building-Executables) for Android documents.

Example:
```python
version = 0.78
```
or even a dynamic version would probably be better:
```python
version = str(self.data.version)
```

jbnimble:
Turns out the latest version of Kivy and Buildozer no longer use the main.py "version" variable and instead use the buildozer.spec "version" property.

BarbourSmith:
👍 👍 👍 

Let me know if you get it to build, it's on my todo list but a ways off 😀

jbnimble:
@BarbourSmith, Built fine, updated wiki instructions. Able too execute on older Android tablet and opened various menus. Did not try to load gcode, and no hardware to test serial connection yet. Thinking of writing a script that takes care of editing the buildozer.spec and running the buildozer commands. Where could a script pull the version number?

PR #492, opened 2017-12-09
Include Sprocket Geometry in Triangular Kinematics in GC Simulation

reecej:
Updated the chain length calculations for triangular kinematics in GC simulator to include sprocket geometry effects.

Accounted for chain length wrapped around the sprocket as well as difference in distance of the chain on the sprocket compared to the motor axes.

Additional discussion, as well as an explanation of the geometry involved can be found at: https://forums.maslowcnc.com/t/accounting-for-motor-sprocket-induced-error-when-calculating-chain-lengths/1422

This is the corresponding simulator update for the triangular kinematics update in firmware: [PR #337](https://github.com/MaslowCNC/Firmware/pull/337)

### DIFF
--- a/Simulation/kinematics.py
+++ b/Simulation/kinematics.py
@@ -75,6 +75,10 @@ class Kinematics():
     Lambda = 0
     Gamma = 0
 
+    # Motor axes length to the bit for triangular kinematics
+    Motor1Distance = 0#left motor axis distance to sled
+    Motor2Distance = 0#right motor axis distance to sled
+
     # output = chain lengths measured from 12 o'clock
     Chain1  = 0#left chain length
     Chain2  = 0#right chain length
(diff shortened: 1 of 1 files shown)


blurfl:
That's great! We'll have to make sure folks update both GC and Firmware together.

blurfl:
I'm finding that GC crashes when I choose Triangular kinematics and click 'Recompute'. Rolling back to the release version of kinematics.py works, so I think the divil is in these details. Haven't had time to look deeper.

reecej:
I must have messed something up. Let me review it.

blurfl:
Is it possible that 'R' isn't defined here?

reecej:
You're right, it should be 'R' rather than 'self.R'. Good eye, thanks for spotting that. I'll correct it.

reecej:
Hmm, hang on a second. I do see 'R' defined here:

    R             = 10.2                                #sprocket radius

And 'self.R' is used elsewhere, in particular in the quadrilateral kinematics. Thoughts?

blurfl:
Two thoughts:
 - self.R is probably OK
 - maybe we shouldn't be using hard-coded values here, but the values from grounbdcontrol.ini instead. That's a different PRm though 😊

reecej:
I agree on the hard-coded values. :-)

I'm suspecting that self.Motor1Distance and self.Motor2Distance should actually be Motor1Distance and Motor2Distance. What do you think?

davidelang:
R needs to be calculated based on the chain pitch and the number of teeth (I am 
told that the firmware is already doing this)

blurfl:
The variables used in the Chainx calculations seem to be present (printed just before the crash):
Motor1Distance 1528.668506   
Motor2Distance 2172.451933  
self.R 10.200000  
self._yCordOfMotor 1072.600000

reecej:
I just pushed an update, would you mind giving it a try? I see that you have MotorxDistance rather than self.MotorxDistance, so I updated the code accordingly.

BarbourSmith:
Looks fantastic! Thank you for being so thorough and taking care of both sides!

PR #493, opened 2017-12-12
Add the ability to change the look of Ground Control

BarbourSmith:
This pull request adds the ability to change the look and feel of ground control. You can now choose between a 'light' and a 'dark' theme.

This pull request is not the final say in how the UI will look, but it puts in place the mechanisms to change the look so we can work towards one we all like...and I want to merge this branch so I can start a new one without an egregious typo in the name 😀


UIElements/screenControls.py
```diff
@@ -2,10 +2,23 @@
 from kivy.uix.popup                            import   Popup
 from UIElements.otherFeatures                  import   OtherFeatures
 from DataStructures.makesmithInitFuncs         import   MakesmithInitFuncs
+from UIElements.buttonTemplate                 import   ButtonTemplate
 
 
 class ScreenControls(FloatLayout, MakesmithInitFuncs):
     
+    
+    def setButtonAppearance(self):
```
BarbourSmith: Is this function used ever?

BarbourSmith: yes, it is

BarbourSmith:
Here is the new UI look in light mode:

image


And dark mode:

image

PR #496, opened 2017-12-15
Change the default encoder steps from 8148 to 8114

BarbourSmith:
As identified in [this ](https://forums.maslowcnc.com/t/what-is-correct-chain-pitch/1430/75)forum thread the correct (or more correct) value for the encoderSteps variable is 8114.

Because this update effects a value stored in the users settings file not only does the default value in ground control need to be changed, but also the users .ini file needs to be updated.

The PR will detect if a user is using the old value and if they are the value will be updated and a popup will let them know that the change has been made.

image


krkeegan:
Should we recommend more strongly a calibration?

I don't really have a good feel for what the error someone would see if they didn't redo the calibration

So they would have a wrong motor distance, but this would be compensated for by the test cuts.  Both of those are abstracted from the steps into dimensions.  I think that is right?  

In which case, changing the steps now probably improves their error, but a new calibration would improve it more?

BarbourSmith:
> In which case, changing the steps now probably improves their error, but a new calibration would improve it more?

I think this is the case. I would expect that everyone will see an increase in accuracy, but the increase will be more if a calibration is done. I'm a little bit worried about giving people the impression that a recalibration is necessary for things to work which could be frustrating

krkeegan:
> which could be frustrating

Yeah, totally agree.  While I think this is important, it doesn't seem like the changes will be earth shattering to most people.

Alright I think your note is probably sufficient.

krkeegan:
Also, we might be able to use a decimal value 8113.7.  I still haven't had the chance to track down where this goes wrong in the firmware.  I should be able to this weekend, but off to Star Wars tonight!

davidelang:
On Fri, 15 Dec 2017, Kevin Robert Keegan wrote:

> Should we recommend more strongly a calibration?

Yes, after this changes, a recalibration is needed.

The errors introduced during the calibration (to counter the other errors in the 
system) will no longer be correct now that the chain is a different length than 
it used to be.

PR #498, opened 2017-12-16
Ensure Machine is Ready; Log Inc Messages Immediately

krkeegan:
There was no check for the 'ok' command from the firmware before sending the first message.  This is likely why PR #405 and commit 6c934aa which sent a blank line on startup helped, becuase only the first command could be sent without an ok.  Removed the kludge of a blank command and it seems to work well.

This may also fix the issues reported in this post:
https://forums.maslowcnc.com/t/unable-to-execute-command-machine-settings-not-yet-received/1517

Also added a new child class of Queue which logs the messages to the logger on put rather than waiting for the process to get them out of the queue. The get method was resulting in non-sequential incoming and outgoing messages in the log because outgoing messages were logged immediately.  This makes the log much more helpful.

### DIFF
--- a/Connection/serialPortThread.py
+++ b/Connection/serialPortThread.py
@@ -39,7 +39,7 @@ class SerialPortThread(MakesmithInitFuncs):
         
         self.bufferSpace       = self.bufferSpace - len(message)
         self.lengthOfLastLineStack.put(len(message))
-        machineIsReadyForData = False
+        self.machineIsReadyForData = False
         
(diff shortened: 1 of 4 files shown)


blurfl:
Both are good solid contributions 💯👍🏻

blurfl:
Does this require the update PR#347 to firmware as well?

BarbourSmith:
I'm reading through that one right now and I will merge it immediately

PR #499, opened 2017-12-16
Properly Track and Change Units on Triangular Test Cut

krkeegan:
The Units button didn't work in the Triangular Test cut page.

### DIFF
--- a/UIElements/measureMachinePopup.py
+++ b/UIElements/measureMachinePopup.py
@@ -462,7 +462,7 @@ class MeasureMachinePopup(GridLayout):
             self.data.message_queue.put("Message: Couldn't make that into a number")
             return
         
-        if self.unitsBtn.text == 'Inches':
+        if self.unitsBtnT.text == 'Inches':
             dist = dist*25.4
         
         dist = 1905 - dist #1905 is expected test spacing in mm. dist is greater than zero if the length is too long, less than zero if if is too short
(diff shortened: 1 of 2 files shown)


blurfl:
Works as advertised, an important fix. 👍

BarbourSmith:
I think I made the same changes in #497 let me double check to see what I missed

UIElements/measureMachinePopup.py
```diff
@@ -492,11 +492,11 @@ def stopCut(self):
         with self.data.gcode_queue.mutex:
             self.data.gcode_queue.queue.clear()
     
-    def switchUnits(self):
-        if self.unitsBtn.text == 'MM':
-            self.unitsBtn.text = 'Inches'
+    def switchUnits(self, id):
+        if id.text == 'MM':
```
BarbourSmith: This is a really elegant way of doing this! Awesome!

BarbourSmith:
It looks to me like changes with the same effect were made in #497. I'm sorry for leaving it unmerged over the weekend. I figured since the issue had been there since the beginning there was no rush, but it looks like I was wrong 😬 and it was a notable issue this weekend.

Great work tracking down what was going on and fixing it, I am sorry for being slow to merge my PR. I will work on adding @blurfl's suggestion that we have a "sanity check" on values entered right away

Issue #494, opened 2017-12-14
Auto Cal Step 10 Triangular version improvements


blurfl:
The instructions for this step would be better if they included the target dimension so the user can do a reality-check on a measurement. 
 Further, the software could do a bounds check anbd refuse values that are wildly wrong. Instead the wild value is written to groundcontrol.ini and the machine will be unable to find a valid position in the workarea.

BarbourSmith:
The distance is added in #495 

I think checking for reasonable values should be done in a pull request which removes the text box where the distance is entered and replaces it with a touch compatible key pad which could do the testing. We already have a keypad widget and adding a "reasonable value" check to it wouldn't be too difficult

blurfl:
Thanks for adding the text 👍. The pop up would be great, I messed around a bit but didn’t see a way to put up a message. 
While you’re in there, the text printed in the triangular routine refers to changing motor spacing instead of radius

krkeegan:
On a related note, there should be a sanity check for the dimension entered by the user.  For example, it shouldn't every calculate a negative radius, which I have seen it do

It should also capture if the user fails to change from MM - Inches.  We don't have to fix it for them, but we should at least give some sort of warning that the value looks very wrong.

BarbourSmith:
Yes! My plan for this was to:

1) Move the calibration process out into it's own object (because having everything in one file was ugly, and we want to be able to run the test pattern without going through the whole process some day)
2) Switch to using the on screen keypad widget to enter the cut distance making it so a keyboard is not required
3) Add a sanity check to the entered value

blurfl:
Wishes have been fulfilled!

PR #501, opened 2017-12-19
Add a sanity check to make sure that the values entered are realistic…

BarbourSmith:
… in triangular calibration

We have had issues with this text box accepting any number. If the number is not a realistic number say (1,000,000) mm the machine will still do it's best to find a calibration which works for that number resulting in nonsense calibration (like a negative rotation radius). This pull request adds a sanity check which limits the entered value to +-300mm of the expected 1905mm distance

I have not had a chance to test this and would like to merge it after the release on Wednesday because I don't want to make too many changes right before the release for fear of bugs slipping through

### DIFF
--- a/UIElements/triangularCalibration.py
+++ b/UIElements/triangularCalibration.py
@@ -71,7 +71,7 @@ class TriangularCalibration(Widget):
         maximumRealisticError = 300
         
         if abs(errorAmt) > maximumRealisticError:
-            self.data.message_queue.put('Message: ' + str(dist) + ' is ' + str(errorAmt) + 'away from the target distance of 1905mm which seems too larget to be real.')
+            self.data.message_queue.put('Message: ' + str(dist) + 'mm is ' + str(errorAmt) + 'mm away from the target distance of 1905mm which seems wrong. This measurement will be ignored.')
         elif abs(errorAmt) < acceptableTolerance:               #if we're fully calibrated
             self.carousel.load_slide(self.carousel.slides[11])
         else:
--- a/UIElements/triangularCalibration.py
+++ b/UIElements/triangularCalibration.py
@@ -62,17 +62,20 @@ class TriangularCalibration(Widget):
         if self.unitsBtnT.text == 'Inches':
             dist = dist*25.4
         
-        dist = 1905 - dist #1905 is expected test spacing in mm. dist is greater than zero if the length is too long, less than zero if if is too short
+        errorAmt = 1905 - dist #1905 is expected test spacing in mm. dist is greater than zero if the length is too long, less than zero if if is too short
         
         print "The error is: "
-        print dist
+        print errorAmt
         
         acceptableTolerance = .5
+        maximumRealisticError = 300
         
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Addresses the second half of #494

BarbourSmith:
Good point! That would require having a different test pattern for different machines because a 1905mm test cut won't fit on a vertically oriented machine which only has 4' of travel in the x-direction. I think it's beyond the scope of this pull request, but it's a good idea

blurfl:
Check spelling 'larger' of the message. The message appears in a floater on the screen, not in the terminal subwindow, yes?

BarbourSmith:
Thanks @blurfl !!! 💯 👍 👍 

That message will show in the regular "Continue" popup. I changed the text so now it reads something like:

"43214mm is 41309mm away from the target distance of 1905mm which seems wrong. This measurement will be ignored."

I'm still not sure that's the most clear wording. Any suggestions for wording could be more clear?

davidelang:
On Tue, 19 Dec 2017, BarbourSmith wrote:

> Good point! That would require having a different test pattern for different 
> machines because a 1905mm test cut won't fit on a vertically oriented machine 
> which only has 4' of travel in the x-direction. I think it's beyond the scope 
> of this pull request, but it's a good idea

This is part of the reason that in the triangular calibration topics I've been 
leaning heavily on having the user move the sled around rather than calculating 
based on expected machine size

blurfl:
@krkeegan is right,  "please check the number and enter it again" would be an improvement 🙂

BarbourSmith:
Great recommendation! I've updated the text to recommend that they try again. We should double check in testing this that it is in fact possible to enter the number again 😉 but from looking at the code it is.

After this is merged, I will make one more change which is to use the keypad object (updated in #503 ) to do the entering of the number and choose the units which I think will be more clear. Then I will test and do the same for the quadrilateral calibration step

blurfl:
the good news - the sanity check works as desired
the not-as-good news 
 - switching the MM/INCHES after typing in the 'Measurement' field does not alter the value displayed. This may be a feature, it aids the case where one typed in a MM value when the setting was INCHES
 - after typing an acceptable value and clicking 'Enter value', that button remains available and if clicked again causes the rotationRadius value in groundcontrol.ini to be reduced by the calculated error again. Can we disable the 'Enter value' button when accepting a value?

krkeegan:
I don't think the MM->inches should convert for the user.  This isn't really a scenario in which they need a conversion.

The enter value reclick could be a bit misleading I suppose and could be triggered by overclicky people. That may be worth disabling.

blurfl:
Disabling is pretty simple - set self.enterValuesT.disabled = True in enterTestPaternValuesTriangular after setting self.cutBtnT.disabled = False, and set self.enterValuesT.disabled to false in stopCut
 Not sure how to do a PR to a PR, though 😊

BarbourSmith:
Excellent use of the review process! 👍 👍 👍 

I've made it so that the enter values button will disable when a value is entered, and re-enable when the test pattern is run again

BarbourSmith:
Great! Let's 🚢

PR #506, opened 2017-12-21
Fix issue where files in light view do not render fully

### LINKED ISSUE #505: New light UI not showing files
I think I found a bug where some files are not shown fully with the new 'light' UI theme. I think the issue is that it is switching back to the color white for some reason

image

BarbourSmith:
There was an issue where some files would not fully display on the light theme because the line color was incorrectly set to white

Fixes #505 

image


BarbourSmith:
It was an easy fix 😁

Issue #507, opened 2017-12-21
Double character in 


blurfl:
PR#503 has introduced an issue where after changing the MM/INCHES in the number input characters typed are doubled.

BarbourSmith:
WEIRD! I would not have expected that. I will look into that right away! Thanks for the catch!

BarbourSmith:
I'm not seeing the same behavior, help me better understand what you are seeing

We're talking about the numbers typed in here showing up twice, right?

![image](https://user-images.githubusercontent.com/9359447/34273778-a4f6586a-e64a-11e7-860a-7b89892dd95f.png)

blurfl:
Right. I open the dialog, type a number, change mm to inches, backspace and type.

BarbourSmith:
I still can't make it happen 🤕

BarbourSmith:
Maybe it's platform dependant?

blurfl:
I've just replicated the issue on Win10 and a second instance of OS X. Seems to be related to clicking in the number field:
1. open the 'machine movement' change window
2. type '12'
3. click in the number field
4. type '34' - see '3344'

BarbourSmith:
Got it! I totally see it too now. I was missing the type in the window field step

BarbourSmith:
I have a proposed solution in #509

PR #509, opened 2017-12-21
A possible fix for issue #507

BarbourSmith:
Issue #507 happens when the text input is selected when entering a distance. This PR fixes the issue by disabling it so that it cannot be selected


BarbourSmith:
Great! Good catch @blurfl, I think that bug has been hiding since we first added the popup!

PR #514, opened 2017-12-22
Ground control osx actions and settings buttons hidden under window title

blurfl:


BarbourSmith:
Thank you so much for tracking this down! People have been mentioning it for weeks, and I looked into it but since the issue doesn't show up on my screen I couldn't tell what was going on. This is exactly the kind of thing that rocks so much about open source. You can see the issue AND you can fix the issue. Awesome! ❤️

Issue #504, opened 2017-12-21
listSerialPorts slow with bluetooth enabled.


madgrizzle:
My laptop has some bluetooth serial ports enabled (not sure why, but COM3 and COM4 are bluetooth.. don't recall ever doing anything with bluetooth serial, nonetheless, they are there).  When I press the actions button, GC calls listSerialPorts and, with my Windows 10 computer, tries to open COM1 through COM256.  When it hits COM3 and COM4, it pauses for seconds... seemingly giving it time to attempt to establish a bluetooth serial connection.

It appears the purpose of opening then immediately closing the port is to determine which of the 256 comports exist (at least for Windows).  However, pyserial has serial.tools.list_ports.comports function that enumerates my COM3 and COM4 into a list.  For linux / osx, is there a particular need to test comports or do ya'll think we can just switch over to using serial.tools.list_ports.comports and not go through the open/close routine?

BarbourSmith:
Great deduction about what is going on! I have a vague recollection that the code you are talking about was added to fix and issue where serial.tools.list_ports was causing a bug...does anyone remember what the issue with using the function was?

blurfl:
How about opening a PR - I can test it on osx and one flavor of linux.

madgrizzle:
Found a stackoverflow thread that describes the issue (GC uses the precise code that was shown there).  However, the bug might have been fixed.

[https://stackoverflow.com/questions/12090503/listing-available-com-ports-with-python](url)

I'll work on a PR to test it.

blurfl:
It looks like serial.tools.list_ports.comports() does work, it builds a list with only the open ports.  I don't have a way to reproduce the slow response though (I'm on a VM, no BLE devices available).

blurfl:
There seem to be a number functions in connectMenu.py that are duplicated in serialPort.py but never used.
 It looks like updatePorts(), and detectCOMports() could be removed from serialPort.py, and the 'win' portion of listSerialPorts() moved into connectMenu.py.
 The 'linux/cygwin' and 'darwin' portions of connectMenu.updatePorts() are one-liners, the 'win' section needs the logic in listSerialPorts()
 Does it make sense to move these serial port discovery routines into connectMenu.py and leave serialPort.py with setPort(), openConnection() and connect()?

BarbourSmith:
>Does it make sense to move these serial port discovery routines into connectMenu.py and leave serialPort.py with setPort(), openConnection() and connect()?

[...]

PR #517, opened 2017-12-23
Clean up connectMenu.py and serialPort.py; use serial.tools.list_ports.comports()

blurfl:
There seem to be a number functions in connectMenu.py that are duplicated in serialPort.py but never used.
It looks like updatePorts(), and detectCOMports() could be removed from serialPort.py, and the 'win' portion of listSerialPorts() moved into connectMenu.py.
The 'linux/cygwin' and 'darwin' portions of connectMenu.updatePorts() are one-liners, the 'win' section needs the logic in listSerialPorts() - instead of making a list of 255 COM ports and testing each one, use serial.tools.list_ports.comports() to create a list of open COM ports

See discussion in issue #504 'listSerialPorts slow with bluetooth enabled'

### DIFF
--- a/Connection/serialPort.py
+++ b/Connection/serialPort.py
@@ -18,7 +18,7 @@ class SerialPort(MakesmithInitFuncs):
     '''
     
     
-    COMports = ListProperty(("Available Ports:", "None"))
+    # COMports = ListProperty(("Available Ports:", "None"))
     
     def __init__(self):
         '''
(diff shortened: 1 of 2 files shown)


BarbourSmith:
This looks great to me. I can't properly test connectivity because I am home visiting my family for Christmas, but I can test that it opens and runs! I'm going to merge it right away because other changes are waiting. Good collaboration!

Issue #518, opened 2017-12-23
GroundControl does not load .NC (with capital letters)


MarmadukeMSB:
Mastercam posts with file extension .NC (capital letters).  Ground Control won't recognize this as an nc file until I manually change the extension to have lowercase letters.

BarbourSmith:
Thanks for making the issue for this so we don't loose track of it. This should be an easy fix 🤞

BarbourSmith:
So after looking into it a little bit I realized that we already have a settings option for which file names work, but for some reason it was disabled. It's under "Ground Control Settings" and looks like this:

![image](https://user-images.githubusercontent.com/9359447/34323475-becf78be-e7ff-11e7-9f77-803cfe181584.png)


I've created PR #519 which re-enables the setting so you can add things to it, and added .NC to the default list of allowed files. Unfortunately, if you already have Ground Control installed and it has already created a settings file it won't be added to your list and you will have to change the setting manually before GC will open those files for you

blurfl:
What is the line I would paste into ground control to enable this?

blurfl:
This change does not seem to have the desired effect. The list of file extensions seems to be controlled by the filters defined in groundcontrol.kv/LoadDialog/FileChooser line 437

BarbourSmith:
I remembered something like. I think that is why the setting is disabled. I will look into it more and try to understand what all is going on. I know I was able to open files with a capital .NC  after adding ".NC" to the list in the Valid File Extension setting.

Maybe it's time to take this issue as a prompting to do a complete re-write of the file opener. Everyone has always hated the Kivy file chooser and we do have another option which is to use the system default file chooser which is accessable through tkinter (https://stackoverflow.com/questions/26607268/simplest-way-to-open-a-file-in-tkinter) which would make opening a file look like every other program. It feels ugly to have both Kivy and Tkinter in the same program so I've been avoiding it, but it seems like it could solve a lot of issues

blurfl:
I've been poking around, and if I could figure out how to give loadDialog.py access to data.config.get() could make the present setup work.
 I agree, though - it would be very nice to replace the Kivy filechooser ! Would adding tkinter affect using buildozer for packaging?

BarbourSmith:
I was wondering the same thing, but my hunch is that it would get bundled in the same way pyserial does. We'd want to test it tho 😉

BarbourSmith:
I believe that this has been fixed with changes to the file chooser 👍 👍

PR #519, opened 2017-12-24
Add capital .NC to files extension

BarbourSmith:
Enables the user setting for file types (why was this disabled again?) and adds .NC to the defaults list

partial fix for #518


BarbourSmith:
I am going to close this for now. #525 would make this un-necessary and it's been open too long and doesn't really solve the issue

PR #525, opened 2017-12-27
Try tkinter file chooser

BarbourSmith:
This branch tests using the tkinter file chooser which will look more like the native file chooser on most operating systems. The kivy one leaves a lot to be desired. It's slow and strange looking the first time users see it. The new system has a lot of pro's and a couple cons:

Pros:
-looks like opening a file in every other program
-much cleaner code (20 lines in one function instead of 70 in three places)
-solves the issue with the file type not being user selectable
-adds a drop down to choose between 'gcode' files and 'all files' which will let you input any file type if your program uses a different extension

Cons:
-Clicking on the main UI while the file chooser is open will hide the file chooser and leave you with a locked up UI. This is especially frustrating because all of the other popups can be closed by clicking outside them on the main UI so it's an intuitive thing to do

I would love feedback on what we think of the new system

image

### DIFF
--- a/UIElements/viewMenu.py
+++ b/UIElements/viewMenu.py
@@ -3,8 +3,10 @@ from   UIElements.loadDialog                     import   LoadDialog
 from   UIElements.pageableTextPopup              import   PageableTextPopup
 from   kivy.uix.popup                            import   Popup
 import re
-from DataStructures.makesmithInitFuncs           import MakesmithInitFuncs
-from os                                          import    path
+from DataStructures.makesmithInitFuncs           import   MakesmithInitFuncs
+from os                                          import   path
+from Tkinter                                     import Tk
+from tkFileDialog                                import   askopenfilename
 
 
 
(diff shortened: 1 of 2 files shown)


blurfl:
>Clicking on the main UI while the file chooser is open will hide the file chooser and leave you with a locked up UI.

This would be a step backward. As a workaround, can we have the function put up a Kivy dialog to cover the main UI with a 'Cancel' button in its center and overlay that with the tkinter file chooser? That way clicking outside the tkinter box would reveal the 'Cancel' button. Have to programmatically close that Cancel screen when the tkinter dialog exits, though.

blurfl:
Struggling to get tkinter to work within groundcontrol...

OSX 10.13.3 Tcl/Tk version 8.6.7 crashes with this error (two different installations):
libc++abi.dylib: terminating with uncaught exception of type NSException
Abort trap: 6
Note that the tkinter._test() works correctly on these machines

BarbourSmith:
Yeah, I agree that we shouldn't merge it until we have a fix for the locking up the UI issue. I'm going to see if I can make the popup automatically close if the popup loses focus 

Are you seeing those issues on OS X when trying to run the build or even when running from the source?

blurfl:
I see the same issue any way I try to run it. I'm looking for more info and a fix...

BarbourSmith:
Switching to file browser is another alternative:

https://github.com/kivy-garden/garden.filebrowser

blurfl:
I _really_ would like to get the tkinter version going. Let me poke at it a little longer?

BarbourSmith:
Me too...I'm still poking at it also 🙂

blurfl:
I'm going to admit defeat on this one. Does it work on Hannah's MacBook? I've got access to two more 'virgin' machines if the problem is local to my installations here, but hesitate to borrow them.

BarbourSmith:
I'm kind of close to admitting defeat on the front of making it not lock up the UI also. I posted a question on Stack Overflow so if that doesn't get any response it might be time to admit that using to interface providers (kivy and tkinter) in the same program is a bad idea.

How are we feeling about [file browser](https://github.com/kivy-garden/garden.filebrowser) as an alternative?

BarbourSmith:
Here's what that would look like:

image

blurfl:
That's a definite improvement 👍🏻😉 over the present one.

BarbourSmith:
Great! I'll close this PR and make a new one for that 😀

PR #527, opened 2017-12-29
Switch to the kivy garden file chooser

BarbourSmith:
The new file chooser allows for filters, switching between views, and easier system navigation.

It now looks like this:

image

This needs to be tested on multiple operating systems before merging


BarbourSmith:
Tested good on OS X 10.11.6 and Windows 10

blurfl:
Works a treat in OSX, editing GC settings extensions list works as well. Is there a way to make the extensions list at the bottom un-editable? Editing it doesn't seem to make any difference? 👍💯

BarbourSmith:
Thanks for testing that so quickly! I'll dig into the options and see if I can make that uneditable 😀

BarbourSmith:
So I confirmed that you are right that the filters do not seem to do anything, then read through the docs and issues for the project on GitHub and didn't find any mention of them not working (or any mention of how to disable them) so I had another go at trying to figure out how to make them work. I think the key (pun fully intended) is that you have to press "enter" after making a change for it to apply. The filters are then cleared from the text box. Overall, I think it's counter intuitive and disabling the text box is a better solution since we have a prefered method of changing the list which is to do it in settings.

Here's how it looks now that it is disabled:

image

blurfl:
Dusabled is better. Editing in the GC settings works well enough.

PR #533, opened 2018-01-02
Update frontPage.py to correct inches/mm conversion

madgrizzle:
The code to convert from values on the screen from inches to mm (and vice versa) assumes used 1 in to 25 mm conversion factor.  It's more accurate to use the values defined just above in the code.

Just a little clean up.

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -142,11 +142,11 @@ class FrontPage(Screen, MakesmithInitFuncs):
         
         if newUnits == "INCHES":
             self.data.gcode_queue.put('G20 ')
-            self.moveDistInput.text = str(float(self.moveDistInput.text)/MMTOINCHES)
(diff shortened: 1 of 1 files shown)


BarbourSmith:
I'm getting the issue I was having before we went to 25 as the conversion factor where the conversion is happening with arbitrarily high precision making the text overflow like this:

image

madgrizzle:
Ah, well, I'll see if I can add some rounding to it.  Let's see if I can edit the PR.

BarbourSmith:
Great solution! I think commiting to the same branch again should automatically update the pull request

madgrizzle:
I think it updated.  Python is strange (and new) to me.  As with all things you don't know how to do, [https://stackoverflow.com/questions/455612/limiting-floats-to-two-decimal-points](url) FTW.

BarbourSmith:
Stack exchange is a life saver.

This works great for me. Good fix!

PR #537, opened 2018-01-02
Update the command to wipe the EEPROM

### LINKED ISSUE #532: B07 (Wipe Eeprom) results in -> error: Invalid gcode ID:78
OS: Linux
FW: 1.02
GC:1.02

->Actions -> Advanced -> Wipe EEPROM
Result in GC=
B07
error: Invalid gcode ID:78

BarbourSmith:
Updates to use the new command to wipe the EEPROM. This uses the maximum option of wiping the entire 4k EEPROM

Fixes #532

### DIFF
--- a/UIElements/diagnosticsMenu.py
+++ b/UIElements/diagnosticsMenu.py
@@ -4,7 +4,8 @@ from UIElements.scrollableTextPopup              import    ScrollableTextPopup
 from kivy.uix.popup                              import    Popup
 from UIElements.measureMachinePopup              import    MeasureMachinePopup
 from UIElements.calibrateLengthsPopup            import    CalibrateLengthsPopup
-from Simulation.simulationCanvas                 import SimulationCanvas
+from Simulation.simulationCanvas                 import    SimulationCanvas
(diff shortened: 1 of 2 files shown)


krkeegan:
Looks good.  Do we want to also push the settings currently stored in GC?  Otherwise the user is left in a state where the firmware has different settings than what the user sees in GC.  This rectifies itself on a restart.

BarbourSmith:
Great point! I will add a trigger to push the settings after a wipe. Thanks for catching that 👍 👍

BarbourSmith:
Wiping the full 4k of memory takes long enough that the connection drops and then comes back like this:

image

(ignore the pcb version not being detected, I'm running a cut and testing on an arduino without a shield)

BarbourSmith:
I added a 6 second delay before the settings are pushed to ensure that the eeprom is fully wiped. I think the process is always going to be a little cluttered with at least one prompt to re-calibrate popping up, but I that's OK wiping the EEPROM completely does require a recalibration at least.

I really like the warning that the EEPROM could not be read and the default settings are being used @krkeegan . Nice design.

PR #540, opened 2018-01-03
Deletes the open/close com port routine

madgrizzle:
This PR removes the open/close com port routine as it is currently used only for Windows and is not needed because serial.tools.list_ports.comports() provides a list of existing comports.  Opening/closing the ports results in a delay for computers that have bluetooth serial ports.


blurfl:
This works on OSX 👍

BarbourSmith:
Works on Windows 10! Great work!

PR #542, opened 2018-01-04
Update SetSprocketVertical

madgrizzle:
This PR updates the calibration routine for setting one tooth of each sprocket vertical.  It provides for 0.1, 1, and 5 degree rotations CW and CCW for each sprocket.  The amount of rotation is calculated based upon the number of teeth and the chain pitch, so this should be accurate for those using non-stock hardware.

I've tested this on my development workstation, and am confident it will work, but have not tested it on an actual Maslow as its friggin cold here in Tallahassee, FL.  Seriously, 24 F in my shed..

Also, this is my first attempt to PR using github desktop/atom and such.  So, I'm crossing my fingers on this.


blurfl:
Seems to work as advertised. Haven't checked for accuracy, but the movement commands have what look like the correct values. It will take some testing to see whether the 0.1degree command actually moves 😁

krkeegan:
I am sensing a tread of decreased maslow usage during the cold months.  Perhaps we should add Maslow parkas to the store?

BarbourSmith:
It looks perfect to me and everything test great! Congratulations on your first pull request 👍 👍 🎉 🎆

PR #544, opened 2018-01-05
Cleanup

madgrizzle:
Deleted defs for LeftCW, LeftCCW, RightCW and RightCCW that are handled by setSprocketsVertical.py.

Also, Atom seems to want to 'cleanup' blank lines.  Sorry.. I can't figure out how to stop it from doing so.


BarbourSmith:
Thank you for taking the time to do this. Keeping the code clean and orderly is so important to having something that's easy to work on! 💯 👍 👍

PR #546, opened 2018-01-05
Show default values next to each setting

blurfl:
Omit the settings (COM port Macros and Macro names) where defaults don't make sense. Hard-code the 'Open File' as "your home directory"


BarbourSmith:
Looks fantastic! I'm really glad we found a way to make sure that each variable is only kept in one place 👍 👍

Issue #369, opened 2017-07-21
Suggestion: Allow GroundControl to update arduino


JDogHerman:
Include avrdude into GroundControl allowing the arduino to be updated without the IDE.

Ground control could query the code on the arduino and prompt the user that an update was needed.

BarbourSmith:
Yes! I think this is a great idea.

PR #547, opened 2018-01-07
Update Triangular Kinematics Calibration

reecej:
This updates the triangular kinematics calibration routine. Rather than request the user to perform multiple cuts, a single cut is performed and then the machine parameters are iteratively found based on the measured parameters.

I have not been able to test the G code or final calibration accuracy as I don't have a Maslow yet, so I would request that this be throughly tested before being merged.

More information at: https://forums.maslowcnc.com/t/calibration-pattern-for-triangular-kinematics/879/34


blurfl:
This works slick! 🎉💯. The value reality-check will save much confusion 😉

First time through, a couple comments:
 - How about makeing a short horizontal cut (5mm?) at each mark, it would make it easier to measure and also help those of us using a pen instead of a bit 😊
 - When done cutting, don't park in the middle of the thing I'm trying to measure!! 😉
 - Having the reality check for the input values is good, but 0.8mm is a valid value for a pen... I can work out how to add 3.325mm and pretend to use a 6.35mm bit if you think I'm out of line, though.
 - My first time, I hid my groundcontrol.ini file and started fresh. The Triangular cal calculated 131.9mm, my previous autocal was 128.1. This is a top-mounted linkage so this new value is in the neighborhood. 
 
I did some test measurements:
 - I moved 20" up (measured 24 7/8") and 20" down (measured 20") from the center. The center is within 1mm of the vertical center of the sheet - this may be an issue with my top measurement.
 - I moved 36" left and repeated the 20" up/down. The center mark was vertically correct but 5/32" short of 36". The top mark was 3/64" short of 20" and 1/8" short of 36". The bottom mark was correct vertically but 9/32" short of 36". Return to center hit the center mark spot on.
Those variances may be frame issues, there are things I've got to check.

This is a definite improvement. Thank you for your work, an excellent contribution!

reecej:
Thank you for trying it out and providing your feedback!

- A short horizontal cut at each mark makes a lot of sense, and I'm sure would greatly ease measurements. I'll make that change.
- Good point on not parking it in the middle of the measurement path after!
- I didn't think about pen users. My goal was to ensure a user didn't put 0.25" instead of 6.35mm, but I think it is much more likely for a user to have a pen instead. I'll change that so the minimum allowed is 0mm.

I greatly appreciate your feedback on the accuracy measurements. The calibration routine assumes the horizontal distance between motors is accurate from the earlier calibration steps, so that parameter is not adjusted. Were the accuracy issues you found at 36" left similar to your previous calibration parameters?

blurfl:
Yes,  similar, perhaps smaller now.  I can restore the previous settings and see if there’s a difference.

reecej:
I made updates per your comments. Would you be able to test it again when you're able and see how it works? 0mm bit diameters are now allowed to facilitate pen usage, 5mm horizontal marks are cut at each spot, and the sled should move out of the way after.

I would also be interested in seeing if there is a difference with your previous calibration routine, more to make sure that we at least aren't moving backwards in accuracy.

reecej:
Also, if your errors on the 36" left measurements are similar with both calibration routines, then I may try to add a chain sag compensation next, as it sounds like the error you are seeing is similar to what chain sag might produce.

reecej:
As an aside, I have also developed a chain sag compensation algorithm for the firmware and GC. It would be integrated into this calibration process. Should I incorporate it here, or wait for this to be merged and then create a separate PR?

davidelang:
do it as a separate PR to adjust the chain length calculation for everything.

reecej:
Sounds good to me. I have the firmware code written to account for chain sag. I'll wait for this to be merged to do the GC code, as this code will need to be modified to provide automatic calibration for chain sag.

blurfl:
Thanks for the recent modifications, they're improvements. 👍  We may get some push-back from folks wanting to use inches instead of mm for the measurements...
The accuracy issues on the left edge are very like the values I had with the old style calibration.  I've put the values for two different calibration runs into tables, below. Interesting, I get a different value for rotationRadius today (various - 134.1; 134.6; 135.1) than I got yesterday (131.9). Not sure what I've done differently, but the result seems to remain in this range today. I kept log files and terminal captures in case they were interesting/useful?
 Thanks again for taking up the challenge for this improvement!

screen shot 2018-01-07 at 6 00 11 pm

I've got some sag measurements as well:
screen shot 2018-01-07 at 6 02 41 pm

reecej:
Thank you for gathering that information, it is very helpful! Would you be able to send me the log files for me to review? My e-mail address is listed in my GitHub profile. The sag measurements are extremely helpful as well, and it looks like sag correction could go a long way to improving the horizontal errors you observed.

Regarding the changing rotation radius values, that could make sense, in particular if we are having sag-induced error of even a small amount. Do you happen to recall which values you measured for the distance between cuts?

I was wondering about the metric vs imperial units. In the previous triangular calibration it seemed only metric units were allowed, whereas on the quadrilateral calibration the user could select. Do you think it is worthwhile allowing the user to select metric or imperial?

davidelang:
Yes, it is a good idea to support both unit systems.

davidelang:
note that the distances for the "sag over X" are diagonal distances, not horizontal distances

davidelang:
a note on measuring with a tape measure.

The reason the hook on the end of a tape measure is loose is to that it moves by the thickness of the hook.

That way, if you hook it over something, you get the same measurement that you would get if there was a vertical surface there that you were pushing the end of the tape against.

cutting a short horizontal mark instead of just drilling a hole makes it much easier to tell where the top and bottom edges are (make the cut at least 1/2" wide, probably better to make it 1" wide so that even large tape measures can hook in the slot)

always measure from the bottom of one slot to the bottom of the other slot (or top to top), that way it doesn't matter how large a bit was used to make the slot.

when allowing entry of imperial units, see if you can support entering fractions.

UIElements/triangularCalibration.py
```diff
@@ -25,104 +28,177 @@ def cutTestPaternTriangular(self):
...
+            # If we get unrealistic values, reset and try again with smaller steps
+
+            if (motorYcoordEst < -(workspaceHeight/4) or rotationRadiusEst < -100):
+                    motorYcoordEst = distWorkareaTopToCut + (bitDiameter / 2) + (workspaceHeight / 4)
+                    rotationRadiusEst = 100
+                    motorYcoordCorrectionScale = float(motorYcoordCorrectionScale/2)
+                    rotationRadiusCorrectionScale = float(rotationRadiusCorrectionScale/2)
+                    print "Estimated values out of range, trying again with smaller steps"
+
+        if n == numberOfIterations:
```
davidelang: at least as a debug option, we should output both the number of iterations, and the accuracy we think we are at.

reecej:
I can make an update to make the slots 1 inch wide, to facilitate easier measurements. Imperial units shouldn't be difficult to do either. Native support for fractions are a different story, and may be something to be included later!

davidelang:
could you do this in firmware as a B (or $) code and have GC invoke that?

This way the chain sag work will directly help everything, not just the 
calibration step.

reecej:
The chain sag will be a complete implementation, including:

- New firmware/GC triangular kinematics algorithms (written)
- New firmware/GC configuration parameters (written)
- Additions to the triangular calibration to calculate the required chain sag compensation (waiting for this PR)

blurfl:
> make the slots 1 inch wide
Something like 10mm or half an inch would be wide enough, don't want to make bigger marks than we need 😉. The sled doesn't need to travel quite so far out of the way, and perhaps should go back to 0,0 when finished?
I see the calculations in the log file, but I don't think I saw the values initially entered. Would those be worth logging?

davidelang:
I'm asking if you can make it so that the iteritive calculations can be done in 
the firmware rather than in GC

This would need to be a series of two commands

1. set location 1 (tell the firmware to remember the chain lengths and current Y 
position')

2. set location 2 (tell the firmware to remember the chain lengths and current Y

3. actual distances are X and Y (do the calculations and save/report the results


This would ensure that any improvements to the firmware (like chain sag/stretch) 
are automatically used by the calibration

This would also make it possible to do the calibration with any tool, not just 
GC

It makes sense to keep the machine specific math limited to the firmware if we 
can.

davidelang:
On Mon, 8 Jan 2018, Scott Smith wrote:

>> make the slots 1 inch wide
> Something like 10mm or half an inch would be wide enough, don't want to make bigger marks than we need

I intially thought of saying 1/2" wide, but then I realized that 1" is a pretty 
normal width for tape measures when you get to the longer ones (25+ ft), it 
takes very little time to do this, and so better to support as much as possible 
rather than get into a state where the marks aren't wide enough for the hook of 
the tape measure to fit.

reecej:
> I'm asking if you can make it so that the iteritive calculations can be done in
the firmware rather than in GC

This is possible, but not with how the current code is written. I like the idea, but it would require that the kinematics formulas be written such that calibration patterns can be used. The alternative would be to rewrite the existing values, but with them stored in EEPROM we have to worry about write endurance. I was thinking a separate PR would re-write those algorithms so that could be allowed, but it will take some effort.

BarbourSmith:
This is VERY exciting!! As soon as I finish getting caught up on email (sooooo much email), the rest of my day is going to be on testing this idea. It's a brilliant concept and I'm so glad you did this! Great work!!!

reecej:
I appreciate it! I'm still working on some tweaks, per the great feedback in this thread. I have some of the updates done, but some of it will have to wait until tomorrow. So tomorrow expect another commit with some tweaks!

BarbourSmith:
I've just done my first run through the new calibration process and this is so great! Thank you for taking the time to even update the pictures. 

I found a new rotation radius of 143.8mm where my old one was 131.9. 

How important is the accuracy of the vertical motor offset measurement to this process? I didn't do a very careful job on that measurement because in the past it wasn't important. If it is important I think we need a better technique than asking people to kinda eyeball when the chain is at the same height of the top of the sheet.

When I run the 1905mm horizontal test cut after the calibration finishes I get 1896mm

blurfl:
> I found a new rotation radius of 143.8mm where my old one was 131.9

Is this with the ring linkage: As the true value for that linkage is easily measured, it's a good test of the calculation done in this method. It sounds like some correction is needed, yes?

BarbourSmith:
Yes, this is with the ring...but I would really like to check for bowing in my frame before I am considered the ideal test case

reecej:
Great question. The vertical motor offset measurement in step 5 of the calibration process is still needed, but an approximation is all that is required. The only thing this measurement is used for is to try to ensure the sled is on the workspace when cutting the measurement pattern. It could be replaced with a generic value, but in an effort to keep the machine as adaptable as possible I was envisioning leaving it in there to help avoid problems with non-standard designs in the future.

Regarding your horizontal test cut distance, is this typical or not? It sounds like @blurfl has experienced similar horizontal inaccuracies. My next PR for chain sag will hopefully improve that error, but it will need to be tested also.

Next commit coming shortly!

BarbourSmith:
Fantastic! Then it worked great. I just wanted to make sure a sub mm measurement wasn't necessary in step 5.

I've been finding that my horizontal cuts are dead on using the old calibration system, but that my vertical cuts (especially near the top) are coming out slightly off

reecej:
That is expected I believe. With chain sag unaccounted for, the accurate dimension will be the one that is measured. That is why I have been planning to follow this up with the chain sag PR as quickly as possible, so hopefully by the next release chain sag is incorporated and users don't have to worry about this.

davidelang:
On Tue, 9 Jan 2018, reecej wrote:

> Great question. The vertical motor offset measurement in step 5 of the 
> calibration process is still needed, but an approximation is all that is 
> required. The only thing this measurement is used for is to try to ensure the 
> sled is on the workspace when cutting the measurement pattern. It could be 
> replaced with a generic value, but in an effort to keep the machine as 
> adaptable as possible I was envisioning leaving it in there to help avoid 
> problems with non-standard designs in the future.

As long as the two cuts are on the sheet, the initial yoffset value is never 
needed.


I would like to propose that instead of trying to move to coordinates for the 
calibration, we instead let the user do the movement

1. we know the distance between the motors, run both chains out to 1/2 of 
that.

   The sled will be very high (possibly off the top of the sheet), but we 
will know that it's not off the bottom. IF it's off the top, feed out a little 
more (say 60% of the motor spacing instead of 50%)

2. let the user run the sled down to where it's on the sheet. Make the first 
mark

3. let the user run the sled down to another point (and tell them that the 
further apart the two marks are, the more accurage the result)

4. make the second mark

5. move the sled to the side (or have the user do it)

6. do the measurements and enter them.


This can all be done without having anything entered for machine dimensions, it 
also will work for standard 4x8 machines, 8x4 machines, 5x5 machines, 2x4 
machines, etc.

davidelang:
On Tue, 9 Jan 2018, BarbourSmith wrote:

> I found a new rotation radius of 143.8mm where my old one was 131.9.

What should it be based on the circle radius and the bearings? that's almost 
1/2" difference.

blurfl:
" let the user run the sled down"
The present approach is more automated, and takes the work area dimensions into consideration. It too will work on all those machine dimensions.

reecej:
Additional commit made, per feedback from here. Please let me know what you think!

reecej:
It's time for a noob question. I see there is a conflict in groundcontrol.kv. If I resolve that conflict, will pressing the "Commit Merge" button on that screen just resolve the conflict, or will it merge the entire PR? I wanted to resolve the conflict, but was afraid of accidentally merging the PR itself.

blurfl:
If you're working on a branch of your fork, I don't think you can merge to the Maslow master branch by mistake. I'm no git expert, though...

davidelang:
If you are working in your repository, you cannot accidently merge into the main 
one.

Even if you did manage to do that, it's still possible to recover and undo 
things.

reecej:
Gave it a shot but I'm still seeing the conflict. I'll await support from more experienced folk. :-)

BarbourSmith:
There was a conflict where we (you and I) both edited the same line of code. It has to do with an update that I made to the UI where the text boxes are gone and instead the number entry popup is used which let's it work on touch screens without a keyboard. I would say don't worry about it and I'll add back any features which are removed accidently after the pull request is merged.

So what is next for this idea? I think it's absolutely the way we want to go forwards.

I think we have two options

- Option 1: We wait to find out why the result isn't perfect before merging this system. Having errors for unknown reasons can come back to bite us in the butt later
- Option 2: We know this is the direction we want to move in and we merge right away to give ourselves a base to build off of.

My vote is that we do some more tests to get a good comparison and then merge now.

To really compare the two systems I think we need to come up with a comparison test that we can repeat to determine if we're getting better as we go along. Here's my proposal for a test pattern:

image

The four marks will let us get accurate big measurements of overall lengths and the five rectangles will give us accurate measurements with calipers in 5 locations on the board.

What are everyone's thoughts on how to proceed?

reecej:
So, it really depends on how you guys want to proceed. The current calibration routine had error on the vertical plane, as it calibrated the horizontal. This calibrates the vertical, pushing the error to the horizontal.

The real value of this is to facilitate simpler calibration of chain sag without requiring the user to weigh their sled. If we just want to jump to that phase as part of this PR, we can do that too.

nereo1:
I like the test pattern you propose -- Maybe align the outside edges of the 4 corner rectangles with the 4 straight grooves? You could do multiple test passes on 1 sheet by moving in 1/2" or 1" for each successive test

davidelang:
On Wed, 10 Jan 2018, reecej wrote:

> The real value of this is to facilitate simpler calibration of chain sag 
> without requiring the user to weigh their sled. If we just want to jump to 
> that phase as part of this PR, we can do that too.

how can you calibrate chain sag from this?

davidelang:
On Wed, 10 Jan 2018, reecej wrote:

> This calibrates the vertical, pushing the error to the horizontal.

The current calibration isn't really measuring anything, it's just modifying one 
value to introduce errors to counter other errors in the system.

The new one is actually trying to calculate the right answer for the geometry.

I think we need to do tests, but the first thing is to figure out why we are not 
getting the expected results.

for the top mount kit, the arms are 5" long (7" for a few) and we add a master 
link (and for some people a half link) to the chain, which actually starts 1/2 
link past the 12 o'clock position on the sprocket, so we should end up very 
close to 5.375" or 5.675" for the rotation radius

the 45 kit is 9" plus the chain attachment

If we are off on these, what is being missed? is it imprecision in the motor 
spacing? or something wrong with the sprocket compensation?

As a test, can someone do short tests with a caliper in several places on the 
board.

Up near the top where the chains are at a shallow angle so the sprocket 
compensation is minimized

near the middle

down near the bottom where the chains are at a steeper angle.

and if someone is doing a 4' wide x8' tall  machine, can you test very low.

Ideally, try and measure chain sag in these different areas so we can calculate 
what the effect would be.

does the calibration result in the same radius offset in all three places?

how much do the differ?

what is the theoretical radius for the kit in question?

blurfl:
" we add a master link (and for some people a half link) to the chain, which actually starts 1/2 link past the 12 o'clock position on the sprocket"
If the half link is positioned on the king tooth, it is counted as the first link of the chain. It is the same length between pins as the other links, it just has one wide end and one narrow one. This is how I do the calibration and was used for the results I posted.

davidelang:
On Wed, 10 Jan 2018, nereo1 wrote:

> You could do multiple test passes on 1 sheet by moving in 1/2" or 1" for each 
> successive test

move the workpiece, don't offset the test.

One mistake that we have made with the calibration is trying to be clever and 
ofsetting the tests in software. the chain lengths/tension/sag change as you 
move to the side, so it adds additional error to the testing.

I didn't know enough when I introduced that misfeature early on. :-(

blurfl:
"move the workpiece" - that's one reason I argue against starting from the center of the workpiece or doing a 'victory lap'.

davidelang:
On Wed, 10 Jan 2018, Scott Smith wrote:

> "move the workpiece" - that's one reason I argue against starting from the center of the workpiece or doing a 'victory lap'.

you need to start in the center, otherwise the math get too hard (the triangles 
on each side are not even)

going to two exteme corners when you are done with the basic calibration, if not 
a full trip around the perimeter is a good practice to find where the limits of 
your movement are.

davidelang:
On Wed, 10 Jan 2018, BarbourSmith wrote:

>> what is the theoretical radius for the kit in question?
>
> I will need to model all the parts accurately in OnShape including the 
> material thicknesses of the production parts to get a 100% accurate 
> measurement of what the correct radius for the ring is, but here is a quick 
> version.
>
> The radius of the ring itself is 105mm. From the ring to where the chain 
> attaches is about 35.4mm. This is taking into account the 1/2 link which is 
> not counted initially.

does this account for how much the link is moved by the pin you use to hold it 
in place? I doubt you use up exactly 1/2 of the link

> This gives us a total expected radius of right around `140.4`
>
> From my test yesterday:
>
>> I found a new rotation radius of 143.8mm where my old one was 131.9.
>
> So it looks like the new system is already beating the old system hands down! Very exciting!

when you re-do the calibration, can you measure the chain sag, that will tell us 
how much it's affecting things at that point.

BarbourSmith:
> does this account for how much the link is moved by the pin you use to hold it
in place? I doubt you use up exactly 1/2 of the link

Not exactly, but it's my best approximation. It also doesn't account exactly for the thickness of the metal parts, but it's as close as I could get trying to worm my calipers in there. I would say I could be off by as much as 3mm so the new system could be dead on

BarbourSmith:
Here is the test pattern I am going to run:

image

[test pattern.zip](https://github.com/MaslowCNC/GroundControl/files/1620505/test.pattern.zip)

Any tweaks we should make before I start running it?

blurfl:
140.4 is about a half a link away from 143.8 which the auto cal calculated. The chain cal starts with the middle of a link at 12 o'clock. @BarbourSmith, How does the chain attach to the circular linkage? Could this be the difference?

BarbourSmith:
Here's a photo of the attachment:

image

You can see that most of the link is taken up, but possibly not the full 1/2 link that the sprocket offsets by

Off topic: have you guys seen the guy who made a darn sailboat in the forums!? No need to respond and create a tangent but it is worth checking out immediately

nereo1:
Any tweaks we should make before I start running it?
Maybe make the test pattern approx. 4' x 6' instead of 8', so the sheet can be slid over for multiple tests?

BarbourSmith:
Sorry my picture was misleading, I zoomed in too much so we couldn't see the edges of the sheet. Here's a more zoomed out version:

image

reecej:
> how can you calibrate chain sag from this?

I will incorporate a pair of cuts in the horizontal plane to account for chain sag, so it will become four total cuts. The lengths of the two chains for those cuts will be used to add chain sag, as well as compensate for any vertical drift of the sled for the horizontal cuts.

Also note that without chain sag compensation, it's unlikely that this algorithm will get the completely accurate rotation radius value. It should be within a couple mm, but chain sag would still throw it a bit off.

Thanks everyone for your feedback, and I look forward to seeing the results of the test cuts!

davidelang:
@BarbourSmith I would add a full horizontal cut across the bottom so you could check for bowing there.

davidelang:
On Wed, 10 Jan 2018, reecej wrote:

>> how can you calibrate chain sag from this?
>
> I will incorporate a pair of cuts in the horizontal plane to account for chain sag, so it will become four total cuts. The lengths of the two chains for those cuts will be used to add chain sag, as well as compensate for any vertical drift of the sled for the horizontal cuts.

could you explain the logic behind this?

I was trying to figure out a way to do something, but trying to measure an error 
of <2mm across 8' with enough accuracy to compensate for it was something I 
never did figure out how to do

> Also note that without chain sag compensation, it's unlikely that this algorithm will get the completely accurate rotation radius value.
> It should be within a couple mm, but chain sag would still throw it a bit off.

I agree, but I think that we can get 'close enough' by having a default sled 
weight and calculating based on that. As the sled weight will vary, it will be 
off by a bit, but I'll bet that it's going to be off a lot less with sag 
compensation and the weight off by 50% than with no compensation at all.

And we should be able to get the sled weight to within 5 pounds or so by having 
someone stand on a bathroom scale, get their weight, and then pick up the sled 
and get their weight again.

That's starting to get pretty close.

reecej:
Absolutely. The way my proposed algorithm works is to begin by getting the rotation radius and motor Y offset to approximately correct values. The vertical cuts (the ones the calibration process cuts per this PR) are used for that. Then, compare the user measured horizontal distance between the new cuts and compare that to the original calculated distance. The user measured distance should be shorter. @blurfl provided a great example measurement, where after calibrating the vertical cuts he moved left 36" of center and measured a horizontal distance of approximately 35.75". The actual horizontal cuts would be from right edge-ish to left edge-ish, so say double the 36". In this example he would likely measure the 72" theoretical cut as 71.5". Granted, 12.7mm is not a lot of error to account for, but the algorithm in this PR runs until an accuracy of 0.001mm is achieved, so it would still compensate for it.

I agree that we could likely use generic values for the weight of the sled, chain weight, workspace angle, etc. Honestly, part of the enjoyment for me is simply the challenge. So the idea of making an algorithm that uses say two measurements to determine the rotation radius as well as compensate for sled weight/chain weight/workspace angle/etc is just a fun problem to solve. I also think it would provide a pretty neat user experience, as well as future adaptability. But I agree that it isn't necessary.

BarbourSmith:
I just finished running the first set of tests and the data is bad. I am not sure why. I almost just re-did it without posting, but I think it could be valuable information. I know that something went wrong because I did the 1905 mm horizontal test yesterday and got 

> When I run the 1905mm horizontal test cut after the calibration finishes I get 1896mm

This time I got 1797mm so something went really wrong. I switched GC branches several times and worked on other things so who knows what happened. Here's the data (although it doesn't mean anything 🙄 )

image

From now on I will start with a fresh settings file, do the calibration process end to end, then run the test immediately after finishing the calibration.

reecej:
Great information, and thanks for sharing. It is very value.

What machine parameters were you using for these cuts?

reecej:
Also, didn't your Gcode intend the cut out area to be smaller than the total work area size?

reecej:
In the Gcode file you attached, the horizontal cut distance is 1805mm, giving 71.06". This makes sense for your measured 1797mm.

davidelang:
On Wed, 10 Jan 2018, reecej wrote:

> Absolutely. The way my proposed algorithm works is to begin by getting the 
> rotation radius and motor Y offset to approximately correct values. The 
> vertical cuts (the ones the calibration process cuts per this PR) are used for 
> that. Then, compare the user measured horizontal distance between the new cuts 
> and compare that to the original calculated distance. The user measured 
> distance should be shorter. @blurfl provided a great example measurement, 
> where after calibrating the vertical cuts he moved left 36" of center and 
> measured a horizontal distance of approximately 35.75". The actual horizontal 
> cuts would be from right edge-ish to left edge-ish, so say double the 36". In 
> this example he would likely measure the 72" theoretical cut as 71.5". 
> Granted, 12.7mm is not a lot of error to account for, but the algorithm in 
> this PR runs until an accuracy of 0.001mm is achieved, so it would still 
> compensate for it.

I'm not sure we will end up with that much error.

We had someone (I don't remember who now) measure the chain sag when they were 
near the lower left, and the measurement was something like 38mm over >3m of 
chain, which I calculated to be an error of ~1mm in the length of the chain.

So I'm expecting that once we have things properly calibrated, we should be 
seeing an error of ~2mm over the 72" width.

I hope you are right in that there's a significant amount of error, as that will 
be far easier to measure and correct for.

BarbourSmith:
Great catch! @reecej You are absolutely right that the issue was in the gcode...great catch!

I have modified the gcode to use the normal length.

What do you mean by "didn't your Gcode intend the cut out area to be smaller than the total work area size?" ?

blurfl:
The sag error is affected by the direction that the sled is moving. The numbers I gave the other day were measured after a vertical sled movement. They might be different after a horizontal movement toward the edge.

BarbourSmith:
Here is the updated gcode

[Test Shape Take 2.zip](https://github.com/MaslowCNC/GroundControl/files/1620794/Test.Shape.Take.2.zip)

reecej:
Thanks @BarbourSmith! Looking forward to seeing the results!

I think Bar's tests here will be informative on the topic. From what we have been seeing, after calibrating the rotation radius using the vertical calibration routine in this PR we see that a cut across the horizontal axis is commonly around 10mm short (1905mm Gcode vs 1896mm sled movement, 1805mm Gcode vs 1797mm sled movement, etc). I have also read of user reports recently of similar error. So I'm suspecting that something in the range of 10-12 mm of error along the entire horizontal axis at Y=0 is realistic.

reecej:
> What do you mean by "didn't your Gcode intend the cut out area to be smaller than the total work area size?" ?

This is me being dumb. Feel free to ignore. :-D

BarbourSmith:
With the gcode issue fixed we're back to 1896 so I think the data is OK here's the final numbers for test 1

image

davidelang:
@BarbourSmith 

did you check for frame flex?

test1 is the new calibration, correct?

can you also measure the outside of the squares at teh top and bottom (is the top wider or narrower than 1896 that we get at the middle?)

at some point, we need to get someone to spend some time with the simulator and see what measurement errors would result in this.

BarbourSmith:
I am using the stock frame design. I have not checked for frame flex, but I will. What was the recommended technique?

This is for the new calibration system.

You are asking for me to check that the measurement is the same at the top and bottom of the squares? I can do that.

BarbourSmith:
All of the squares are the same at the top, middle, and bottom to within ~.2mm which is about the limit of what I think is reasonable to consider accurate the measurement

blurfl:
> What was the recommended technique

Measure motor-to-motor before pulling the chain and while the chain is pulled. It doesn't matter what you measure so long as you measure the same two points both times - is there any difference, and how much.

davidelang:
On Wed, 10 Jan 2018, BarbourSmith wrote:

> I am using the stock frame design. I have not checked for frame flex, but I will. What was the recommended technique?

measure with a tape measure with everything is loose, then measure again while 
the chain measuring motor distance is under the max tension.

> You are asking for me to check that the measurement is the same at the top and bottom of the squares? I can do that.

top and bottom of the sheet, not top and bottom of individual squares.

davidelang:
sorry, I was meaning the left edge of the top left square to the right edge of 
the top right square.

that should be the same as the distance measured across at Y=0

simiarly, what is the distance across the bottom two (left edge of the bottom 
left square to the right edge of the bottom right square)

BarbourSmith:
Excellent suggestion.

Updated with those measurements:

image

blurfl:
Dopn't we expect 1805mm across the middle?

BarbourSmith:
Yes, I think we do

The 1896 measurement was after adjusting the gcode to expect 1905 so that it would match with my measurement from yesterday to confirm that something was deeply wrong

reecej:
Yes, the expect 1808 across the middle is actually expect 1805.

reecej:
Thanks for taking the time to do this @BarbourSmith! This a very useful info. I would be curious to have you (and anyone else that is interested) to try out the chain sag correction code that I have as a next step so we can compare to this result. I think that will give us additional insight as to whether the error we're still seeing is sag-induced or due to some other aspect.

BarbourSmith:
I would love to test it. I'm really liking this test pattern, I think it gives us a lot of good information to work from.

I'm working my way through the original calibration process right now and really missing that one click calibration 😉

BarbourSmith:
Just so I can better understand how it works, why does the new calibration ask for a measurement from the cut to the top of the sheet?

reecej:
That is to precisely determine the motor Y offset based on the updated motor Y coordinate from the calibration process. It's the more accurate value of step 5.

blurfl:
> the more accurate value of step 5

Step5 has always bothered me- trying to estimate by eye a distance in a different plane and several inches aside. This is _so_ much nicer!

BarbourSmith:
Here's the results for the current calibration system:

image

BarbourSmith:
Looking at them side by side I think it's pretty much immediately clear that @reecej has made a MAJOR improvement.

Here is both in a single post for easy side by side comparison:
The old system:
image
The new system:
image

reecej:
Great testing, thank you for putting the time in for that!

I suspect the updated calibration method is getting values closer to the actual geometry of the rotation radius, so on average the numbers look better than before. However, this does show that we are trading some horizontal accuracy for vertical accuracy. I'm looking forward to seeing your tests with the chain sag compensation to see if we can get it even closer!

BarbourSmith:
Here is one last test where I did the normal calibration process for the first part and then skipped the calibration process and entered my expected value of 140.4

image

This test pattern is a good metric I think. I am more than happy to run it any time there is a request.

I vote we are ready to merge the pull request and make this the official calibration procedure...anyone opposed?

davidelang:
the fact that the horizontal values get smaller as we go down tells me that some of the parameters are not right (either the rotation radius or motor spacing)

reecej:
I think the horizontal values getting smaller as he went down correlates well with chain sag.

I took Bar's measurements and put them into my chain sag algorithm. I had to make some rough guesses of what measurements Bar would have gotten for the chain sag calibration process, but based on those guesses here are my calculated chain sag errors for Bar's setup:

Y=450mm
1800mm nominal cut is actually 1798.1391mm
1.8609mm too short

Y=0mm
1800mm nominal cut is actually 1794.6549mm
5.3451mm too short

Y=-450mm
1800mm nominal cut is actually 1790.4618mm
9.5382mm too short

These values seem to correlate well with Bar's observed measurements:

Top measurement: 3mm too short
Middle measurement: 6.5mm too short
Bottom measurement: 10mm too short

BarbourSmith:
Fantastic PR @reecej 👍 👍 💯

reecej:
FYI, GroundControl [PR 558](https://github.com/MaslowCNC/GroundControl/pull/558) and Firmware [PR 369](https://github.com/MaslowCNC/Firmware/pull/369/files) have been created as the next step of this.

davidelang:
On Thu, 11 Jan 2018, reecej wrote:

> I think the horizontal values getting smaller as he went down correlates well with chain sag.
>
> I took Bar's measurements and put them into my chain sag algorithm. I had to make some rough guesses of what measurements Bar would have gotten for the chain sag calibration process, but based on those guesses here are my calculated chain sag errors for Bar's setup:

what is your chain sag algorithm?

I've been using the calculations here: 
https://www.spaceagecontrol.com/calccabm.htm?F=100&a=4.407&q=0.13&g=9.81&Submit+Button=Calculate 
and they've seemed plausable

> Y=450mm
> 1800mm nominal cut is actually 1798.1391mm
> 1.8609mm too short
>
> Y=0mm
> 1800mm nominal cut is actually 1794.6549mm
> 5.3451mm too short
>
> Y=-450mm
> 1800mm nominal cut is actually 1790.4618mm
> 9.5382mm too short
>
> These values seem to correlate well with Bar's observed measurements:
>
> Top measurement: 3mm too short
> Middle measurement: 6.5mm too short
> Bottom measurement: 10mm too short
>
>

reecej:
> https://www.spaceagecontrol.com/calccabm.htm?F=100&a=4.407&q=0.13&g=9.81&Submit+Button=Calculate

I used the same source, although designed my own formula based on it in which I consolidated all coefficients into a single variable (chainSagCorrection).

PR #552, opened 2018-01-09
Disable command to zero the axes in Step1 'Begin' button

blurfl:
I think the issue I was seeing was caused by the 'Begin' button in Step1 - it executes gcode "B06 L0 R0" which tells both axes that they are at the first link of the chain. There is a comment "This prevents strange behavior when rotating each sprocket to 12:00".
With the present code, clicking this button clobbers your chain calibration whether that was what you intended or not.
 I believe recent changes to kinematics and the chain calibration procedure make this 'Begin' button behaviour unnecessary.

### DIFF
--- a/UIElements/measureMachinePopup.py
+++ b/UIElements/measureMachinePopup.py
@@ -158,7 +158,7 @@ class MeasureMachinePopup(GridLayout):
         each sprocket to 12:00
 
         '''
-        self.data.gcode_queue.put("B06 L0 R0 ");
+        # self.data.gcode_queue.put("B06 L0 R0 ");
         self.carousel.load_next()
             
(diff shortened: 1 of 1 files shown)


krkeegan:
I agree.  I would just delete the line entirely.

The sprocket movement commands are all incremental anyways, so there isn't really a chance for a big jump anymore.

blurfl:
I was being cautious, but can do another commit with the line removed and the comment above it amended.

BarbourSmith:
I'm going to be honest I somehow missed that this is a pull request and not an issue. My comment was suggesting a pull request...but clearly you have already made one 😀 

Hannah is out this week so I've got twice as many balls to juggle. This looks perfect. Excellent!

PR #553, opened 2018-01-09
Switch to deque to make sure buffer is freed in the right order

### LINKED ISSUE #551: Issue with order quick commands are sent to machine
Quick commands are sent to the machine immediately which means that their length should be but onto the front of the list of sent command lengths instead of the back.

This may require switching to a [deque](https://docs.python.org/3/library/collections.html#collections.deque)

BarbourSmith:
This pull request fixes an issue where when commands are sent to the machine they are not necessarily processed in the order they are sent, for example a stop command may be executed before the line that was being processed is finished.

This pull request switches to using a deque so that quick commands are added to the other side of the stack to indicate that they will be finished first.

@krkeegan does this address the issue you were seeing?

Fixes #551


krkeegan:
I think it fixes some things.  I think I still see an extra 'ok' on the firmware side so I am trying to track that down.

Can someone with the original issue test this and see if they notice any improvement?

krkeegan:
Oh, I might have it, I think we send a full line with the ~ right?  So it actually looks like ~\r\n?  I think when we catch this in the firmware, we are only catching the ~ and leaving the \r\n, which then looks like an empty line that we process.

BarbourSmith:
@krkeegan YES! That makes so much sense to me! You are a genius!!!

It's so subtle, but it makes perfect sense. There is an extra invisible line left in the buffer which is processed resulting in an extra 'ok' being sent. Removing the 'ok' from the '~' like I did in the other pull request fixes the issue because one 'ok' is still sent...but my pull request is far from the right way to fix it.

I can replicate the issue if we need to test anything.

...you are brilliant!!!

krkeegan:
It looks like we also add a space

davidelang:
also, be aware that \r\n is what you get from windows, from Linux or Mac (or 
android or iOS) you will just get \n

this is arguably a bug in GC that it uses the default line ending for the OS 
it's running on instead of the interoperability standard

I'm pretty sure the firmware just ignores the \r, but make sure you don't 
accidently do something that requires it.

krkeegan:
It looks like we add the line ending ourselves and it is set to \n, so I think we are good there.  I think the reason some people see \r\n is that the writing to the log uses the system line endings, which is probably correct.  

It looks like this is a combination of factors.  I am also seeing my hated _watchDog function running requesting more code.  This may be my justification to get rid of that thing.

BarbourSmith:
> This may be my justification to get rid of that thing.

Let's do it. I almost got rid of it in my PR.

krkeegan:
OK, yup this works.  I have one commit to add into this branch and a separate PR for the firmware.

BarbourSmith:
Great, do you want to make a pull request against this branch for the extra commit? We're really getting fancy with the git then 😀

BarbourSmith:
Oh, I see you are way ahead of me 👍

BarbourSmith:
All good to 🚢 ?

krkeegan:
yes

PR #554, opened 2018-01-10
Don't Add Extra Space onto Quick Commands

krkeegan:
I don't think they are needed for regular commands either, but
they only add one more thing that has to be disposed of when
processing a quick command in the firmware.


krkeegan:
This PR merges into your branch, not the master branch.

BarbourSmith:
Skillfully done!

PR #555, opened 2018-01-10
Sleep Serial Thread for Massive Decrease in CPU Usage

krkeegan:
Not sure how we missed this, but I noticed that my laptop sounded awfully loud.  Upon closer inspection python was eating up a full core.  It looks like the issue is we just needed a slight sleep in the SerialRead thread.  I just picked a millisecond, which took my CPU time down to nearly zero.  We could increase this if others with slower machines are noticing issues.

I wonder if this will let me increase my Kivy framerate?


BarbourSmith:
Yesssss!!! Great catch 👍 👍 👍

PR #558, opened 2018-01-11
Chain Sag Correction for Triangular Kinematics

reecej:
This incorporates chain sag correction into the triangular kinematics algorithm. Changes are made to the simulator to align with the associated firmware changes.

Additionally, a new parameter called chainSagCorrection is introduced. The triangular calibration process is updated to determine this value and send to the firmware.

This is the Ground Control PR associated with Firmware [PR 369](https://github.com/MaslowCNC/Firmware/pull/369).


reecej:
I would request @BarbourSmith and others test this thoroughly before being merged. I have run simulations on it and all seems well, but testing on an actual Maslow is the real test.

BarbourSmith:
Man, you are really getting things done @reecej! Fantastic work!

I will run the same test we were running yesterday using this branch and the compirable branch for the firmware. I will start with a fresh settings file and run through the calibration process before starting the test.

BarbourSmith:
I got to the end of the calibration process, but then I ran into an error. When I entered my values I got a warning which looked like this:

image

```
57, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4979, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4980, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4981, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4982, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4983, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4984, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4985, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4986, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4987, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4988, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4989, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4990, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4991, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4992, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4993, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4994, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4995, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4996, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4997, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4998, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 4999, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
N: 5000, Motor Spacing: 2975.34, Motor Y Offset: 429.51, Rotation Disk Radius: 144.057, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: -0.0, Chain Error Cut 2: -0.0, Chain Error Cut 3: 19.1736, Chain Error Cut 4: 3.2198, Sled Drift Error: 0.0
Machine parameters could not be determined
Sending: G21
Sending: G90
Sending: G40
Sending: G0 X0 Y0
```

reecej:
Could you send me the full logs please? And do you know which values you entered for the measurements?

I should modify the calibration process to include the entered measurements as well for troubleshooting.

BarbourSmith:
Yes here is the log:

[log.txt](https://github.com/MaslowCNC/GroundControl/files/1624224/log.txt)

The full output of the terminal is unfortunately lost because it will only let me scroll back so far before the terminal itself stops keeping track. Those same repeating lines scrolled by for maybe a minute before it got to the end

Here are the measurements I entered:

image

reecej:
Were you starting with a clean configuration? Based on the screenshot of the console output it calibrated for the motorYoffset and rotationRadius, and was attempting to calibrate for chain sag but calculated that a negative chain sag would be required. The algorithm doesn't allow this, so it timed out.

BarbourSmith:
I did start with a clean configuration (deleted the Ground Control.ini file)

What would cause it to come up with a negative number?

reecej:
In this case, it looks like the horizontal measurement you had (1861mm) is longer than it expected. For an 8' wide work area it should have tried to cut out a 1830mm wide mark. With a smaller-than-actual rotationDiskRadius setting it would be reasonable to expect the cut out mark to be longer than 1830, but this value was long enough that the chains to cuts 3 and 4 were calculated as being too long rather than too short (i.e. chain stretch, for instance).

Could you confirm you measured from the same side of each mark (left/top/etc) to remove the size of the bit from the equation? The only place the bit diameter is factored in is the motor Y offset. If everything is good there then I have an idea to see if we can make it work.

BarbourSmith:
I am still measuring pretty much exactly 1861mm from the right edge of the left cut to the right edge of the right cut

reecej:
Could you check your settings and ensure the chainSagCorrection value is 0? If so, try rolling back to the previous firmware and repeating the cut to see if the measurement changes. I'm wondering if something is off in the firmware.

BarbourSmith:
I am seeing zero...but I am also seeing that it is in the quadrilateral kinematics mode:

image

I'm 99.9% sure I clicked triangular during the calibration process...no wait I know exactly what happened...I forgot to turn on the z-axis because the switch wasn't working the calibration option (now fixed) and going back probably messed something up. 203% my fault...although if I was in quadrilateral mode how did I get to the triangular calibration window?

reecej:
Definitely not 203% your fault. If you, admittedly very experienced in using the machine, was able to encounter an issue then I am confident a new user could have the exact same issue.

Did you by change press the next step button from step 9 rather than pressing the triangular kinematics button?

BarbourSmith:
It's possible, but I thought I had a lockout that would prevent someone from being able to skip into the wrong one. It should skip you into the correct calibration window.

I am going to try again being very careful from step 1

BarbourSmith:
Good call on asking me to check on the settings! We never would have found it if you hadn't throught to look there 👍 👍

reecej:
I have a theory as to what happened. I see in the code that if no rotationRadius was able to be obtained, then a value of 260 is applied. I'm wondering if you deleted your .ini file if this value would be used instead. The algorithm is written to search upwards for a rotation radius as there are actually two sets of solutions (the motor being below the workspace technically satisfies the math), so with the rotation radius default being so much above the actual value I could see it being confused.

It might be worthwhile having the calibration routine clear the rotation radius and chain sag correction values before calibration if this is the case, to avoid these types of problems.

Thoughts?

BarbourSmith:
Oh good point!

Something that has been on my todo list for a while now is to re-think the order of the calibration process. There is the step where we make a rough estimate of the distance between the points where the brackets attach to the sled and I think doing something similar for the rotation radius would be a good idea.

I am mid way through the calibration process (after deleting the .ini), and I just checked the newly created .ini file and the radius is at 100 right now. I will keep an eye on it through the rest of the process

BarbourSmith:
Here are my settings going into running the test cut:

```
[Maslow Settings]
bedheight = 1219.2
openfile = 
motoroffsety = 432.92
sledwidth = 310
bedwidth = 2438.4
comport = COM4
macro1_title = Macro 1
sledheight = 139
macro2 = 
macro1 = 
zaxis = 1
sledcg = 79
macro2_title = Macro 2
colorscheme = Light
zdistperrot = 3.17
motorspacingx = 2974.83

[Advanced Settings]
digits = 4
kinematicstype = Triangular
rotationradius = 100
kppos = 1300
encodersteps = 8113.73
zencodersteps = 7560.0
enablepospidvalues = 0
kpv = 7
homex = 0.0
homey = 0.0
propweight = 1
enablevpidvalues = 0
kiv = 0
gearteeth = 10
zaxisauto = 0
kipos = 0
kdpos = 34
truncate = 0
kdv = 0.28
chainsagcorrection = 0
chainpitch = 6.35

[Ground Control Settings]
zoomout = pagedown
centercanvasonresize = 0
zoomin = pageup
validextensions = .nc, .ngc, .text, .gcode


```

davidelang:
if you start your calculations at rotation radius =0 and search from there, how 
many iterations before it gets in the ballpark of the right value?

Even if it's 1000, just add another thousand or two iterations to the limit and 
everything's good.

That's far better than having a hard-coded minimum size that calibration breaks 
if we drop below it.

reecej:
The calculations do start at rotationRadius = 0, but it's more of the possibility of the test cut being done with a rotation radius that was substantially larger than the real value. Regardless, even if we clear the rotationRadius on the machine and iterate from complete scratch, like you mention it will take a negligible amount more time to complete the calculations.

BarbourSmith:
I'm getting spookily similar values to last time, should I press "Enter Values"?

image

reecej:
Try it out!

BarbourSmith:
I'm seeing the same thing as before 😐 

This error:

```
N: 4994, Motor Spacing: 2974.83, Motor Y Offset: 425.17, Rotation Disk Radius: 145.15, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.4743, Chain Error Cut 4: 2.9178, Sled Drift Error: 0.0
N: 4995, Motor Spacing: 2974.83, Motor Y Offset: 425.17, Rotation Disk Radius: 145.15, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.4743, Chain Error Cut 4: 2.9178, Sled Drift Error: 0.0
N: 4996, Motor Spacing: 2974.83, Motor Y Offset: 425.17, Rotation Disk Radius: 145.15, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.4743, Chain Error Cut 4: 2.9178, Sled Drift Error: 0.0
N: 4997, Motor Spacing: 2974.83, Motor Y Offset: 425.17, Rotation Disk Radius: 145.15, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.4743, Chain Error Cut 4: 2.9178, Sled Drift Error: 0.0
N: 4998, Motor Spacing: 2974.83, Motor Y Offset: 425.17, Rotation Disk Radius: 145.15, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.4743, Chain Error Cut 4: 2.9178, Sled Drift Error: 0.0
N: 4999, Motor Spacing: 2974.83, Motor Y Offset: 425.17, Rotation Disk Radius: 145.15, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.4743, Chain Error Cut 4: 2.9178, Sled Drift Error: 0.0
N: 5000, Motor Spacing: 2974.83, Motor Y Offset: 425.17, Rotation Disk Radius: 145.15, Chain Sag Correction Value: 0.0
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 19.4743, Chain Error Cut 4: 2.9178, Sled Drift Error: 0.0
Machine parameters could not be determined
```
Scrolling by very quickly and then ultimately the popup say that the values could not be determined. 

The fact that we got the exact same measurements feels like a little bit of a win...at least the calibration process is very repeatable

reecej:
Could you try setting the rotationRadius to 0 and repeating the cut to see what the measurements are by chance?

reecej:
Something seems off in the cut pattern. In the testing you did yesterday the long horizontal cuts were always a couple mm shorter than the nominal distance. Here, it's 30+mm too long. Another option would be to reset your rotationRadius to the proper value and perform the test cut too.

davidelang:
should we do the vertical spacing test, get the results, then do the horizontal 
ones?

k

BarbourSmith:
I agree that its very strange that all the cuts used to tend to short, but now we're getting long. I will run the 1905mm test cut again and see what I get with the current settings

reecej:
> should we do the vertical spacing test, get the results, then do the horizontal
ones?

That's effectively how the algorithm works, but the part that is giving us trouble is moving to that second phase of calibrating the horizontal.

BarbourSmith:
Where is the gcode which is being sent to the machine? For some reason I can't see it in the "files changed of the pull request"

reecej:
It's integrated into triangularCalibration.py. I can get you the output directly though, one moment.

reecej:
[Triangular Calibration Cut Pattern.nc.txt](https://github.com/MaslowCNC/GroundControl/files/1624480/Triangular.Calibration.Cut.Pattern.nc.txt)

BarbourSmith:
Looking through my original log file it looks like the line sent is `G0 X-927.1 Y317.5` so the expected distance would be 1854.2mm so we're not like WAY off

reecej:
Yes, that's to align to the left side after making a cut to the right. A couple lines down you should see the X1828.8, which is the critical measurement.

BarbourSmith:
Oh, got it. The first one is just to position

reecej:
Any luck with that 1905mm cut? I'm beginning to suspect gremlins in the firmware.

It's acting somewhat like a large chain sag compensation value is set in the firmware, however in all your files it's 0. Strange.

BarbourSmith:
I'm getting 1938.5mm for the 1905mm test.

What if I flashed the master branch of firmware and ran the same test again right now with the same GC settings. That would test if it's in the firmware, right?

reecej:
It would. I say it's a good idea!

reecej:
For reference to help understand the console output from the calibration routine, here is how it generally works.

The algorithm begins by trying to calibrate the motorYoffset and rotationRadius parameters using the chain lengths from cuts 1 and 2. So the first thing you'll see is the errors for chain 1 and 2 go to zero, which we see here. Then, we look to the error on cut 4 to calibrate the chain sag. If everything is working right we should see a negative value for the cut 4 error, indicating the chains for that cut location were shorter than expected. But we're actually seeing positive values, indicating the chains were longer than expected.

If all is working then the final calibration step would be to use the chain lengths from cut 3 to determine how much the sled drifted vertically while trying to move horizontally.

If something is off in the firmware then this could be the cause.

As an aside, the firmware you're using is the corrected version with the semi-colons present in the calibration equations, right?

krkeegan:
Keep in mind.  You can open up the serial console in the arduino ide and send $$ to the machine to see a list of the settings and their values, if you think something is being garbled.

reecej:
> Keep in mind. You can open up the serial console in the arduino ide and send $$ to the machine to see a list of the settings and their values, if you think something is being garbled.

Great advice! Thank you! @BarbourSmith, would it be possible for you to try that and see what you get?

blurfl:
or set a macro to $$
Macros, another great tool!

BarbourSmith:
The 1905mm test is running right now with the master firmware and GC. I had to go to master on GC also because it was getting strange issues when I didn't update both

BarbourSmith:
I'm getting 1938 again so there doesn't seem to be something dramatically wrong with the firmware, it's just these settings are off

reecej:
What settings are on GC, versus those in the firmware?

BarbourSmith:
From the log:

```
Sent: $$ 
<Idle,MPos:1050.09,505.53,3.01,WPos:0.000,0.000,0.000>

[PE:-0.00,0.01,127]

$0=2438.40 (machine width, mm, $K)

$1=1219.20 (machine height, mm, $K)

$2=2974.83 (motor distance, mm, $K)

$3=432.92 (motor height, mm, $K)

$4=310.00 (sled width, mm, $K)

$5=139.00 (sled height, mm, $K)

$6=79.00 (sled cg, mm, $K)

$7=2 (Kinematics Type 1=Quadrilateral, 2=Triangular, $K)

$8=100.00 (rotation radius, mm, $K)

$9=2000 (axis idle before detach, ms)

$11=1650 (original chain length, mm)

$12=8113.73 (main steps per revolution, $K)

$13=63.50 (distance / rotation, mm, $K)

$15=1000 (max feed, mm/min, $K)

$16=1 (Auto Z Axis, 1 = Yes)

$17=0 (auto z axis)

$18=12.60 (max z axis RPM)

$19=3.17 (z axis distance / rotation)

$20=7560.00 (z axis steps per revolution)

$21=1300.00 (main Kp Pos)

$22=0.00 (main Ki Pos)

$23=34.00 (main Kd Pos)

$24=1.00 (main Pos proportional weight)

$25=7.00 (main Kp Velocity)

$26=0.00 (main Ki Velocity)

$27=0.28 (main Kd Velocity)

$28=1.00 (main Velocity proportional weight)

$29=1300.00 (z axis Kp Pos)

$30=0.00 (z axis Ki Pos)

$31=34.00 (z axis Kd Pos)

$32=1.00 (z axis Pos proportional weight)

$33=7.00 (z axis Kp Velocity)

$34=0.00 (z axis Ki Velocity)

$35=0.28 (z axis Kd Velocity)

$36=1.00 (z axis Velocity proportional weight)
```

BarbourSmith:
I think the line of note is this one: `$8=100.00 (rotation radius, mm, $K)`

So somehow the rotation radius is not being set

reecej:
This is with the previous firmware, or the PR version?

BarbourSmith:
I am currently running master branch on both GC and firmware, but my last calibration was done using PR558 (this) and the firmware to go with it.

Here's a question which might be silly, but when should the radius be set?

If we start with a default value of 100 and then go through the whole calibration process up until right before pressing the "run test cut" button it shouldn't be changed yet, right?

As soon as I press that button both the vertical and horizontal cuts are made, right? So the algorithm knows what the right rotation radius is by the time it gets to using the numbers from the horizontal measurement, but the cuts are already in the wood from when the old radius was used, right?

reecej:
That's correct. The idea behind it is that the algorithm searches for the parameters which would make the predicted cut dimensions for a given chain length equal the measured dimensions.

The radius should be set after the calibration iteration algorithm completes successfully. One points to note is that during the test cuts the actual chain lengths are never pulled from the firmware, but rather the kinematics formulas are used to replicate the chain lengths. The idea is they should be the exact same kinematics formulas as what the firmware has so they should produce the same result. But issues could also arise if somehow the parameters changed between the cut and the iterative algorithm executing (which they shouldn't be able to).

If you never completed the calibration process successfully in this PR, then the machine parameters should not have been updated.

Does that answer your question?

reecej:
Could you set your rotationRadius to an appropriate value for your linkage kit (140.5 I believe?), then try the new firmware with your 1905mm cut and see what you get?

BarbourSmith:
Ok, I think I'm getting it

The issue we're seeing is that the expected cut length of 

> A couple lines down you should see the X1828.8, which is the critical measurement.

Is coming out at

> I am still measuring pretty much exactly 1861mm from the right edge of the left cut to the right edge of the right cut

But we know that the reason that is happening is that when that cut is made, the machine THINKS the radius of rotation is 100mm...right? 

What I think is that we need a two step process. First we make the vertical cuts, use them and the excellent code we merged yesterday to find the rotation radius. We push these new settings to the machine and THEN make the second set of horizontal cuts which should now be close to correct.

BarbourSmith:
> Could you set your rotationRadius to an appropriate value for your linkage kit (140.5 I believe?), then try the new firmware with your 1905mm cut and see what you get?

Yes!

BarbourSmith:
With the algorithms computed radius we're coming out at 1895mm on the 1905 test which meshes pretty darn well with what you were predicting, right?

reecej:
The difficulty with the two step process is that the chain sag affects the measurement of the rotationRadius, so we'll still be compounding error on top of error.

1895mm is reassuring, so at least the firmware isn't doing anything strange. If you run the calibration process with that rotationRadius, does it work?

I'm suspecting that the algorithm may need some tweaking.

BarbourSmith:
Great suggestion. I will try going through the calibration process starting with my rotation radius at 144

reecej:
Would you mind giving the commit I just pushed a try? Part of my algorithm accounts for the vertical drift the sled moves during the horizontal movement, but the previous algorithm only allowed upwards drift. The feedback you provided appeared to show downwards drift, so I modified it accordingly. With your inputs I was now having the calibration routine complete.

BarbourSmith:
Here is what I got when I started the calibration process with a radius of 144

image

About to press the button and run the new algorithm as of a few minutes ago...

Woo! Here's what I got for final values:

image


```
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0004, Sled Drift Error: 1.5364
N: 736, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.579, Chain Sag Correction Value: 77.797631
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0003, Sled Drift Error: 1.5364
N: 737, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.579, Chain Sag Correction Value: 77.798991
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0002, Sled Drift Error: 1.5364
N: 738, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.579, Chain Sag Correction Value: 77.799788
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5364
N: 739, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.579, Chain Sag Correction Value: 77.800406
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5364
N: 740, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.579, Chain Sag Correction Value: 77.800886
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5364
N: 741, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.579, Chain Sag Correction Value: 77.801089
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5364
N: 742, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.579, Chain Sag Correction Value: 77.801303
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0027, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5364
N: 743, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.801502
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0028, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5364
N: 744, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.801529
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0027, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5364
N: 745, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.801612
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0027, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5364
N: 746, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.801714
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0027, Chain Error Cut 4: 0.0, Sled Drift Error: 1.5364
N: 747, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.801688
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0003, Sled Drift Error: 1.5371
N: 748, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.803134
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0002, Sled Drift Error: 1.5371
N: 749, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.804149
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5371
N: 750, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.804743
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5371
N: 751, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.805204
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5371
N: 752, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.805562
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5371
N: 753, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.805714
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5371
N: 754, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.805873
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5371
N: 755, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.806022
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0021, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5371
N: 756, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.806041
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.002, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5371
N: 757, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.806103
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.002, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5371
N: 758, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.806179
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.002, Chain Error Cut 4: 0.0, Sled Drift Error: 1.5371
N: 759, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.80616
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0016, Chain Error Cut 4: -0.0002, Sled Drift Error: 1.5376
N: 760, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.807238
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0016, Chain Error Cut 4: -0.0002, Sled Drift Error: 1.5376
N: 761, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.807995
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0016, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5376
N: 762, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.808438
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0016, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5376
N: 763, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.808782
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0015, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5376
N: 764, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.809049
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0016, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5376
N: 765, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.809162
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0015, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5376
N: 766, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.80928
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0015, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5376
N: 767, Motor Spacing: 2974.42, Motor Y Offset: 427.848, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.809391
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0015, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5376
N: 768, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.809406
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0015, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5376
N: 769, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.809452
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0015, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5376
N: 770, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.809509
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0015, Chain Error Cut 4: 0.0, Sled Drift Error: 1.5376
N: 771, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.809495
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0012, Chain Error Cut 4: -0.0002, Sled Drift Error: 1.5379
N: 772, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.810299
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0012, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5379
N: 773, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.810863
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0012, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5379
N: 774, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811193
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0012, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5379
N: 775, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811449
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0012, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5379
N: 776, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811649
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0001, Chain Error Cut 3: 0.0012, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5379
N: 777, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811733
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0011, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5379
N: 778, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811821
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0011, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5379
N: 779, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811904
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0011, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5379
N: 780, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811915
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0011, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5379
N: 781, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811949
  Chain Error Cut 1: -0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0011, Chain Error Cut 4: -0.0, Sled Drift Error: 1.5379
N: 782, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811992
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0011, Chain Error Cut 4: 0.0, Sled Drift Error: 1.5379
N: 783, Motor Spacing: 2974.42, Motor Y Offset: 427.847, Rotation Disk Radius: 143.578, Chain Sag Correction Value: 77.811981
  Chain Error Cut 1: 0.0, Chain Error Cut 2: 0.0, Chain Error Cut 3: 0.0009, Chain Error Cut 4: -0.0001, Sled Drift Error: 1.5382
Machine parameters found:
Motor Spacing: 2974.42, Motor Y Offset: 427.8, Rotation Disk Radius: 143.6, Chain Sag Correction Value: 77.81258
Sending: $16=1
Sending: $12=8113.73
Sending: $13=63.5
Sending: $19=3.17
Sending: $20=7560.0
Sending: $21=1300
Sending: $22=0
Sending: $23=34
Sending: $24=1
Sending: $25=7
Sending: $26=0
Sending: $27=0.28
Sending: $28=1.0
Sending: $29=1300
Sending: $30=0
Sending: $31=34
Sending: $32=1
Sending: $33=7
Sending: $34=0
Sending: $35=0.28
Sending: $36=1.0
Sending: $17=0
Sending: $0=2438.4
Sending: $1=1219.2
Sending: $2=2974.42
Sending: $3=427.8
Sending: $4=310
Sending: $5=139
Sending: $6=79
Sending: $7=2
Sending: $8=143.6
Sending: $37=77.81258
Sending: $K
Sending: G21
Sending: G90
Sending: G40
Sending: G0 X0 Y0




```

BarbourSmith:
Should I run the test pattern?

reecej:
Yes! I would be very interested to see how that test pattern compares.

reecej:
Then my only remaining question would be, does the calibration routine work with a brand new .ini file. In this case it looks like your logs showed positive sled drift (look for the sled drift error value in the iterative output), so I would want to ensure that it works with a base configuration which may show negative drift error.

BarbourSmith:
Here's where we're at:

image

Looks pretty darn good to me.
For reference, here's the last test from yesterday:

image

> Then my only remaining question would be, does the calibration routine work with a brand new .ini file. 

I agree that is the question, I will have a go at it first thing tomorrow 😀 

Great work today everyone! It's astounding how much gets done every day.

reecej:
That looks like good progress! Thank you very much for your assistance, I appreciate you sticking in there and helping get this working!

davidelang:
On Fri, 12 Jan 2018, reecej wrote:

> That's correct. The idea behind it is that the algorithm searches for the 
> parameters which would make the predicted cut dimensions for a given chain 
> length equal the measured dimensions.
>
> The radius should be set after the calibration iteration algorithm completes 
> successfully. One points to note is that during the test cuts the actual chain 
> lengths are never pulled from the firmware, but rather the kinematics formulas 
> are used to replicate the chain lengths. The idea is they should be the exact 
> same kinematics formulas as what the firmware has so they should produce the 
> same result. But issues could also arise if somehow the parameters changed 
> between the cut and the iterative algorithm executing (which they shouldn't be 
> able to).
>
> If you never completed the calibration process successfully in this PR, then 
> the machine parameters should not have been updated.

The problem is that without knowing the right rotation radius, when doing the 
horizontal cuts, it thinks it needs to let out more chain than it actually does, 
so the cut is too far out.

This is why setting the rotation radius to 144 first works.

If the actual rotation radius is less than what the machine thinks it is when 
doing the cut pattern, it will cut horizontal distances that are less than 
expected (because if it thinks that the links are 200mm when they are really 
only 100 mm, the machine doesn't let out enough chain)

But if the actual rotation radius is MORE than what the machine thinks it is, it 
will cut horizontal distances that are MORE than expected (if it thinks the 
links are 100mm when they are really 200mm, it lets out 100mm too little chain)

davidelang:
that last pattern looks like it's over-compensating for something, since the distances get wider as you go down.

I think the problem is that the calculations are being thrown off by the incorrect rotation radius that I describe above. The sag calibration is assuming that all the horizontal error is due to the sag, when some of it is due to the rotation radius being off.

In the loop to calculate things, you will need to add a factor for the expected rotation radius from the start of the loop (somehow)

reecej:
With the original algorithm, when I repeated Bar's steps (wipe out .ini file, copy Bar's first set of inputs), I got the same result of not being able to calibrate the machine. With the updated algorithm, I repeated the test and was able to calibrate the machine, so I suspect when Bar tries tomorrow with a fresh configuration it will work now. I've also been making some tweaks to make it more stable, with positive results so far when mirroring Bar's inputs.

I agree, it does look like it is over-compensating doesn't it. In looking through the logs with the current algorithm, adjusting only the rotationRadius and motorYcoord gets the vertical measurement within .01mm and the horizontal measurement within 2mm. However, as the horizontal measurement gets closer to 0mm error the rotationRadius actually goes higher than the expected value. In the logs I'm looking at with Bar's calibration, the algorithm actually takes the rotationRadius nearly to 150mm (and still increasing as the horizontal cut error moves closer to 0). When the chainSagCorrection value begins to be modified the rotationRadius actually starts to come down closer to what I would say the correct value is.

There is the ability to weigh different parameters, in order to have the algorithm prefer to use certain parameters to solve the error. I tried reducing the chain sag influence and increase the rotationRadius influence, but the same end result occurred.

One thing I did notice was the effect of the motor spacing parameter. If the frame exhibits flex during the motor spacing phase of the calibration, it could result in a motor spacing value which is less than the actual measurement. This would cause both the rotationRadius and the chainSagCorrection values to increase beyond their ideal values, resulting in excessive chain sag compensation such as what we're seeing here. I'm wondering if Bar's frame is experiencing this.

davidelang:
what about my thought that the error from the prior rotation radius is causing 
the cuts to be too far left and right?

reecej:
The prior rotation radius could cause the cuts to be further right or left, but the idea behind this algorithm is that it doesn't matter. All we really require are cuts on the workspace correlated to known chain distances, and then we can calculate the machine parameters from those. The problem we were having earlier with Bar's calibration failing was because the algorithm was not compensating enough for vertical drift. But with that fixed, you should theoretically be able to start with any machine parameters, so long as they produce cuts that can be measured.

davidelang:
On Fri, 12 Jan 2018, reecej wrote:

> The prior rotation radius could cause the cuts to be further right or left, 
> but the idea behind this algorithm is that it doesn't matter. All we really 
> require are cuts on the workspace correlated to known chain distances, and 
> then we can calculate the machine parameters from those. The problem we were 
> having earlier with Bar's calibration failing was because the algorithm was 
> not compensating enough for vertical drift. But with that fixed, you should 
> theoretically be able to start with any machine parameters, so long as they 
> produce cuts that can be measured.

the prior rotation radius affects the distance from the sprocket to the bit for 
the same amount of chain.

does your calculation include the prior rotation radius in it's calculations?

If not, that's an error.

set the rotation radius to something extreme (say 1000) and see what the 
resulting calibration looks like.

reecej:
The prior rotation radius is included in the calculations, lines 179-182 of triangularCalibration.py.

BarbourSmith:
As I was making dinner I was thinking about how much I like this test pattern as a measurement of accuracy, but it is still a little bit subjective. Looking at each test is a little bit of a reading of the tea leaves. They are looking better and better, but it's hard to say exactly how much better so I came up with a system to give us a number to benchmark each test to.

The benchmark is two numbers like 1.23-3.45 where the first number is the average of the absolute values of the error in the long measurements and the second number is the average of the absolute values of the error in the short (100mm) measurements. The first number tells us about big distortion across the sheet, the second number tells us about localized distortions. A perfect test would give 0.00-0.00.

I thing that the first number is more important, but I could be wrong.

Applying this metric to our tests so far, we get a score of 5.66-1.46 for the old school calibration system we have been using for the last month. The new calibration system without the chain sag compensation comes in at 4.16-1.46 so a measurable improvement. Adding the chain sag compensation gives us a big boost to a score of 2.50-0.89 a huge improvement. The 8mm error on the lowest horizontal measurement is the bulk of what's holding us back there. Overall a fantastic progression!

BarbourSmith:
I was wondering about flex in my frame too. I tried to measure it today, but I wasn't sure I was doing it right. Measuring the distance between the motors under tension and without didn't get me very far because I just couldn't get accurate enough measurements as just one person for the results to be meaningful. I got a piece of light weight string and hung it between my motor mounts (but not so tight that it would stretch) then I measured the change in the position of the low point in the string with the sled under full tension at the top center and removed. I found that the string sagged 25mm lower when the frame was under maximum tension. Using the catinary curve equation I ended up computing that that deflection was caused by something like a 0.1mm movement of the motors which didn't seem right so I figured I was doing it wrong. Can anyone else make better sense of my measurement?

blurfl:
I get the same 0.1mm...
Can you hook a tape measure on the left sprocket with chain on it, and measure to the left edge of the motor shaft of the right motor or to a screw on the right motor mount? A 12' tape should work, and the tape should hook on the links wrapped around the left sprocket. The shaft of the right motor or the motor mount is easier to see than trying to match the chain placement on the left motor. Pick a place on the right motor that won't have chain in the way when the motor measurement is tight. We're looking for a noticeable, measurable change in the measurement when tension changes. Think 5mm-10mm or so. The Simulator predicts less than .1mm horizontal error for a .33mm error in motor spacing. It takes 10mm error to get near 3mm error

reecej:
I agree that a substantial motor spacing error would have to be present to make a noticeable difference. Less than 1mm motor spacing error translates into very small errors on the work area.

The one manner in which this changes is with the chain sag compensation circuit. It's likely that this algorithm will interpret frame flex/motor spacing error as chain sag, and thereby provide a higher chain sag error than what is actually happening. This seems to be the case for your measurements. I was calculating you should ideally get a chainSagCorrection value of somewhere near 50. Higher numbers mean more sag correction, so the value of close to 80 that you got would mean more chain sag correction than required. This matches what we are seeing in your measurements.

The areas where we measured (the horizontal and vertical centerlines of the work area) are within 1mm of the nominal dimensions, so the algorithm is doing what it is supposed to do to find the solution for the data put in. But the fact that we see errors throughout the rest of the area means that either there are other variables at play (motor spacing, chain stretch, etc), or that the formulae for chain sag I'm using aren't accurate.

An alternate option is to have the calibration process cut a test pattern similar to what you did, with long measurements at the left, right, bottom, top, center-horizontal, and center-vertical. The calibration process could then work to minimize total error across all those measurements. Thoughts?

reecej:
Been playing with some numbers. When you have time, would you be able to manually set your rotationRadius to 140.5 and your chainSagCorrection to 38, then run that same test pattern?

BarbourSmith:
Alright, I'm ready to run tests again. First thing I am going to do is:

> When you have time, would you be able to manually set your rotationRadius to 140.5 and your chainSagCorrection to 38, then run that same test pattern?

And then:

> Can you hook a tape measure on the left sprocket with chain on it, and measure to the left edge of the motor shaft of the right motor or to a screw on the right motor mount? 

And then:

> since you have extra #25 roller chain, and we know the weight of that, can you
attach that to the two ends (under a little tension), measure how much it dips
from flat with no tension, and then how much more it dips while under tension.

reecej:
That could be iterated pretty easily. I'm not sure what to do with the results though, as characterizing frame flex into some semi-accurate equation in a universal manner sounds like a pretty daunting task. Maybe it could be a test of frame stiffness along with a warning we could provide the user?

I've been playing with the chain sag calculations, and I feel the vertical error is what is throwing us off. In Bar's measurements without chain sag compensation he measured a vertical cut of nominal length 900mm measure out to 895.5 mm. The problem is, my equations predict that the 4.5mm error in that measurement isn't related to chain sag. There is chain sag, but the amount of sag at the top is nearly identical to the amount of sag at the bottom. So chain sag correction should move the cut to the proper location, but the distance would be the same. This is why the algorithm is compensating by increasing the rotation radius, as that could solve that error.

However, frame flex at the top-center of the workspace, with chain tension at the highest, compared to frame flex with the sled at the bottom could cause that error. So it will be interesting to see Bar's results with the rotation radius and chain sag statically configured!

BarbourSmith:
Here's what I'm seeing for the new test:

image

reecej:
That looks more like what I was hoping to see! It seems the only spot we have major accuracy issues is that middle vertical measurement.

As a follow-up, would you be able to perform a few extra measurements on that middle vertical line, specifically measuring from the center to the bottom cut and the center to the top mark? trying to determine if the error is uniformly distributed across the board or if it's dominantly on one side.

blurfl:
I wonder if the assymetry is related to the direction of travel? if the squares on the right were cut in the other direction...

BarbourSmith:
> measuring from the center to the bottom cut and the center to the top mark

From center to bottom I am reading 449 and from center to top I am reading 440. 

I measured by making a mark at zero and then measuring from the bottom edge of the middle cut to the top edge of the lower cut and from the top edge of the middle cut to the bottom edge of the top cut (basically the space in between the two cuts) so the bit width is a factor. Good call that the error is not evenly distributed!

BarbourSmith:
> I wonder if the asymmetry is related to the direction of travel? if the squares on the right were cut in the other direction...

I tried to make my gcode so that all of the lines are cut in the same direction. I know that there is some amount of play in my router itself and so I tried to have it always biased the same direction if possible.

reecej:
449mm from the center to the bottom sounds good (compared to the 450mm nominal), so it seems like the only place something is off is at the top center, which also happens to be where chain tension is highest. You don't have a crossbar supporting your motor arms do you?

I suspect we're likely to encounter this with a lot of others, so I'm wondering if the calibration algorithm should be changed to avoid performing measurements in this region.

reecej:
Out of curiosity, is your right motor arm solid, including the motor mount? The error seems to be pointing that direction, so I figured I would completely stab in the dark and mention it. :-)

BarbourSmith:
> Can you hook a tape measure on the left sprocket with chain on it, and measure to the left edge of the motor shaft of the right motor or to a screw on the right motor mount?

After doing this with a decent amount of care I'm getting a reading (measured from some pretty arbitrary points) of 2990mm under no load and 2980mm under load so I am seeing measurable and significant flex

BarbourSmith:
From the "feel" test the right motor feels as strong as the left motor, but that's not much of a metric. It could be that under load the right is moving more. Regardless I think that a top bar style frame is in order!

krkeegan:
> > I wonder if the asymmetry is related to the direction of travel? if the squares on the right were cut in the other direction...

> I tried to make my gcode so that all of the lines are cut in the same direction. I know that there is some amount of play in my router itself and so I tried to have it always biased the same direction if possible.

I am also guessing, but I know one of the errors that I have seen that is hard to correct is the router rotation driving the sled at the lower corners.  This happens because the far chain has very little slack.  

I can't remember the correct rotation of the router, but pretending it is clockwise the router will tend to creep to the right, when descending on the left side and creep to the left when ascending on the right side.  (if the rotation is backwards those descriptions should also be backwards).

reecej:
> Regardless I think that a top bar style frame is in order!

I think this is actually a very good test, as I'm betting more than half of Maslow users are using a frame identical to yours. So this frame flex issue will be a real problem.

My simplest idea to solve this is to modify the calibration algorithm to perform the vertical cut from the center of the work area to the bottom mark, thereby avoiding the top center area. How does that sound? If so, I can submit that to this PR and we can see if it generates better values.

blurfl:
>rotation of the router

I didn't mean that, I meant that when the sled travels toward a motor's edge, the slack chain goes slacker. How does that affect the figures cut and the measurements taken?

davidelang:
@BarbourSmith can you get a 10' 2x4 and screw it across your arms as high up as you can, it's not quite as good as a full top-beam frame, but it will be very close.

The fact that the top center was 440 while the bottom was 440 perfectly matches the frame flex, at the top where the tension is highest, the frame flexes a bit and so the sled sits lower than it should (and 10mm in motor distance error translates to a pretty significant position error)

BarbourSmith:
> My simplest idea to solve this is to modify the calibration algorithm to perform the vertical cut from the center of the work area to the bottom mark, thereby avoiding the top center area. How does that sound?

I think that sounds like a fantastic idea. Way to work within the limitations of the problem! We're installing the tiny house tomorrow so I might not be able to run tests until late in the evening if at all so if I'm off the grid it's not because I'm not excited (I am very excited).

reecej:
Update made to avoid cutting in that top center area. When you're free, would you be able to try deleting your .ini file and repeat the calibration using this update? Thank you!

BarbourSmith:
I most certainly can!

blurfl:
@davidelang 's suggestion 
> can you get a 10' 2x4 and screw it across your arms as high up as you can, it's not quite as good as a full top-beam frame, but it will be very close.

is a good one - there are a lot of folks with 'Standard' frames who will want a simple fix for rigidity, and that would do it. Having it proved in this process would be reassuring for them.

reecej:
I agree, I think this would be great test to perform. It could allow us to put real numbers on the accuracy differences before and after the change. Additionally, it would have a minimal impact on the total project cost and difficulty.

Moreover, later on if something like this is incorporated into the "standard" Maslow frame, then this calibration routine could be modified to incorporate the entire vertical cut as before. This would be helpful, as with the half height vertical cut a 0.5mm measurement error has significantly more impact than it would with the longer cut.

blurfl:
Does a new version of global_variables.py which includes Advanced/_chainSagCorrection need to be part of this PR? My main.py choked until I copied yours settings file across.

reecej:
@blurfl, was there a post from you having issues? Anything I need to look at, or was it resolved?

blurfl:
I had trouble getting it to run until I got a copy of global_variables.py which includes Advanced/_chainSagCorrection. After rereading the exchanges above, I saw you were aware of the situation. 
Haven’t had a chance to use it yet, I hope to today.

reecej:
I look forward to seeing your results. Please let me know of any issues you encounter.

blurfl:
I ran the Triangular calibration in the PR then ran the test pattern on a unistrut top-bar frame with a top-mounted linkage (calibrated rotationRadius = 136.8). The results:

test pattern after chain-sag calibration

reecej:
Thanks @blurfl! Looks like you're seeing too much chain sag compensation also.

Do you recall your measurements you entered from the calibration phase? What is your expected rotation radius, and what chain sag correction value did you get?

blurfl:
The chain sag correction value came in at 138.59301. I mis-quoted the rotationRadius, the value was changed from 134.6 before I started the cal to 138.2 as seen in the settings now. It should be in the neighborhood of 127 to 133.35 on @davidelang 's linkage.

reecej:
138 for chain sag correction I would say is definitely too high. It looks like your horizontal measurement in the center is good, but the vertical measurement got off some how. I'm concerned that performing the vertical calibration over such a small portion of the work area is going to be problematic.

If you guys are willing to try it out, I may switch the calibration process to cut two long horizontal lines instead of one horizontal and one vertical. One would be high on the workspace, and one would be low. I personally want the calibration process to yield better results than this, as it is too inconsistent to be merged in the current state, in my opinion.

blurfl:
I'm willing though it will be tomorrow before I can try it.
 Sure wish the calibration cut locations were indicated somewhere before running - gotta get unmarked stock in place so I can tell which marks are the current run 🙂

reecej:
I'll try to get the changes done tonight, but it might be tomorrow before I get the update in anyway. Thank you for your testing and thorough measurements!

> Sure wish the calibration cut locations were indicated somewhere before running

Any recommendations on how we could help here?

davidelang:
I'm a bit concerned that trying to measure two long horizontal distances is 
going to be hard to do accurately, the differences between the top and bottom of 
the workpiece are going to be very small for a given amount of error compared to 
the vertical error you get from small chain lengths (especially in the upper 
area)
k

davidelang:
On Sun, 14 Jan 2018, reecej wrote:

> I personally want the calibration process to yield better results than this, 
> as it is too inconsistent to be merged in the current state, in my opinion.

I agree with the desire to get better results, but keep in mind that the results 
are already significantly better than stock (even without chain sag 
compensation, it's better than stock)

blurfl:
> Any recommendations on how we could help here?

Something in the text instructions - "the cuts will be centered and about XXmm apart vertically and YY apart horizontally" or so?

blurfl:
> measure two long horizontal distances is going to be hard to do accurately

The sled makes a pretty good clamp to hold the tape measure against the worksheet if you're working alone 😉.

reecej:
I'm more just trying to throw ideas out and see what works best. The new calibration method without chain sag compensation is already merged, so even if this flops that should still be included in the next release!

For this part, my thought process is to try different options and see what works best. So far we have:

Option 1) X=0 full height cut and Y=0 full width cut
Option 2) X=0 half height cut and Y=0 full width cut
Option 3) Two full width cuts (yet to be tested)

If any of those options appears to be a clear winner, based on your guy's testing, I'm all for going for it. Or if it needs to change more, let's do that instead. I figure I'll try writing the code for the two full width cuts, and if it doesn't improve then we can try something else.

Good idea on the description of where the cuts will be. I'll include that too!

blurfl:
I'll be glad to run 'em if you write 'me 😉   Great stuff, much progress, thanks for all your efforts!

reecej:
Update made, so now it uses two long horizontal cuts. I tried it out with the data from @BarbourSmith's and @blurfl's tests with good results.

Give it a test drive and let me know how it goes!

BarbourSmith:
I'm on it!

BarbourSmith:
Here's what I got for calibration values:

image

I did not start from a fresh .ini file

blurfl:
> Give it a test drive

@reecej , Do you want anything zeroed before running this?

reecej:
@blurfl I would think it would be most interesting to start with the default motorYoffset/rotationRadius/sagCorrection values, and see if it can find the right values. It's simpler for the algorithm if the approximately correct values are already entered, so I would say let's test it the hard way if possible!

reecej:
@BarbourSmith nice results! Those look to be much closer to what we would want to see, compared to past versions. Two follow up requests if you have time:

1) Try again with your .ini file cleared, and see if the results are similar
2) Cut one of your test patterns so we can quantitatively characterize the accuracy

Thank you for your work!

BarbourSmith:
Here are my results:

two horizontal tests take 1

For a score of `1.00-1.35` which is amazing!

I am going to try again from a fresh .ini file and see how the score compares. If it's even close to as good, my vote is to merge this today. It's a huge step forwards. I'm sure there are things we can keep doing to improve even further, but I think we've blown way past the "better than what we have now" criteria to merge 👍 👍 👍 👍  (that's both thumbs and both toes up)

BarbourSmith:
Here are the values I got. Very close to before

image

The measurement for the distance between the motors changed by 1.5mm which is concerning. I actually noticed it was different than usual last time. I usually get 2974. Not a part of this pull request, just something to note.

On to running the calibration pattern.

davidelang:
did you put a brace between the wings to stiffen the frame?

reecej:
Great feedback. It looks like if that center vertical measurement is taken out then the first number goes to around 1.4, which is in the realm of what we have been seeing.

I think this is beginning to look pretty good. If @blurfl is still testing today I would be curious to see his results on a different frame, just to ensure no surprises, but otherwise this would likely be a good solution to move forward with.

I also like how this calibration pattern only places cuts in the corners, such that the most usable sections of the board are still usable for other projects.

@BarbourSmith, I think the next step you had of bracing the frame and then remeasuring is a great one. We may be getting to the point where the largest accuracy issue is frame flex, which would be simple to solve.

davidelang:
You can attach a 10' 2x4 with screws or lag bolts so that you can easily remove it.

I think it would be very useful to see the difference that the added stiffness of the frame would add.

At this point, we don't know if the errors we are getting are due to the frame flexing, or due to problems in the calculations/measurements. By reducing the flex, and keeping everything else identical, we can see where we still have calculation/measurement problems.

blurfl:
I've been so thrilled with this new calibration that I keep running it over and over! On my last two runs, the 1-2 measurement has been 1930mm and on this last run the 3-4 measurement is 1930mm as well (down from 1931.2mm on the previous run). The rotationRadius has settled to between 127.7mm and 128.1mm. The chainSag value has varied quite a bit over the runs, from 44.63 with a blank .ini file down to 34.95 then up to 36.1 and up again to 44.63 my last-run-but-one and 29.06 on my last run. Not sure why that would be... 
 At any rate, a tremendous improvement! Thank you @reecej !🎉💯👍
run5 - 1930mm - 1930mm

davidelang:
The fact that different starting conditions result in different chain sag 
results indicates to me that it's not properly accounting for the starting 
rotation radius.

but if you repeat the process a few times, using the results of the prior run as 
the starting point, this should smooth out over a few runs.

reecej:
> The fact that different starting conditions result in different chain sag
results indicates to me that it's not properly accounting for the starting
rotation radius.

If this were the case, then each time you ran the calibration process you would see the rotationRadius increase by 100+mm, even for a calibrated machine. I tried it out, and it was not pretty. But you're welcome to review the code as well. More eyes never hurt.

From my testing, the chain sag correction value is extremely twitchy. Differences of 1mm combined across the two measurements (which could be chalked up to randomness in the sled movement) could cause those types of variations. It will be interesting to see what values users typically get, as well as how much variation we see.

I really appreciate all the testing! It hasn't been the easiest process, but it looks like we're getting somewhere. 👍

blurfl:
That's what the last two runs were, re-runs without changing any values. I've just finished a sixth run with no changes to the .ini file, the cuts were about 0.4mm wider than the previous run, and I got nearly the same values as in the fifth run - chain sag reduced by 0.0016. 
 I'm stoked, gotta get set up for the cal. test file !

davidelang:
On Tue, 16 Jan 2018, reecej wrote:

> From my testing, the chain sag correction value is extremely twitchy. 
> Differences of 1mm combined across the two measurements (which could be 
> chalked up to randomness in the sled movement) could cause those types of 
> variations. It will be interesting to see what values users typically get, as 
> well as how much variation we see.

This is what was stalling me, how to measure to an accuracy of <1mm across 8 ft 
is a hard problem.

here's a thought, can you predict the amount of chain sag at a particular 
position, and then have the user measure it (thread end-to end and then a ruler 
to fine the largest distance between the thread and the chain)

While this isn't suitable for the final calibration routine, it should at least 
be able to tell us how far off we are.

And this deviation from straight is going to be a much larger value (and easier 
to measure) than the difference in length.

reecej:
True, and that goes into my thought process from where we go from here.

So far, I have been grouping remaining error into three categories:
1) Sled "randomness" during movement
2) User measurement error during the calibration process
3) Other influencing physics components that we have not identified/accounted for

My suspicion is that item 1 is contributing a substantial amount to this. As a test, I was designing the next generation of this calibration process (for a separate PR) which would incorporate additional horizontal cuts to use in the calibration process. The goal of this would be to remove some of the "randomness" of sled position, if such randomness exists. But it would substantially change how the calibration algorithm works, so I anticipate that will come sometime in the future.

blurfl:
You've helped the sled randomness issue a lot by the way that you've ordered your cuts in vertical rows. Doing the right side cuts from top to bottom might help it a little as the X-accuracy is better at the top of the work area.

davidelang:
Can someone try doing the calibration a few times, but between doing the 
calibration runs, change the rotation radius to different values (say 0mm, 50mm, 
100mm, 200mm, 300mm) and see what the results look like?

I'm not saying that there's no compensation for the prior rotation radius, just 
that it's not right yet, something isn't been accounted for properly.

I expect that repeated calibration runs with the rotation radius set to the same 
value before the run will result in very similar final numbers, but that there 
will be significant, and consistant differences between the different starting 
values.

reecej:
> Doing the right side cuts from top to bottom might help it a little as the X-accuracy is better at the top of the work area.

Very good point. Think this is something I should modify quickly, or leave as is for the time being?

davidelang:
I don't think the order of the cuts is going to make much difference, as the 
sled moves up and down, it's going to have pleanty of time to adjust any 
side-to-side error.

blurfl:
Do modify it, to have it in this release. Is there  a reason for the horizontal cut not to be connected to cut 2?

blurfl:
@davidelang , the sled makes the cuts in counterclockwise order, so it makes cut 4 after a horizontal travers across the bottom of the aork area. That is sub-optimal for avoiding chain sag error...

blurfl:
> change the rotation radius to different values

I started with rotationRadius at 100, the default, and the procedure set it to 28.1mm, almost perfect. Subsequent runs changed it a half a mm up and down, and it settled on 27.7 on the fourth run.

reecej:
My original thinking for the separation of the horizontal cut was to maximize the side area of cut 2 for measurement purposes. If cut 5 was connected to cut 2, do you think it would make it more difficult to measure?

blurfl:
The cuts seem big enough to not interfere. Spoil less of the sheet, too 😁. I'd cut 1, 3, 5, 2 then 4.

reecej:
Sounds good to me. 1, 3, 5, 2, 4 is the order I have been making the Gcode for right now. I think I'll move 5 to the top edge of 2, so that set of cuts will become a "L" shape rather than a "T" shape.

reecej:
I'm glad I re-visited this, because I did actually find a little error in the Gcode. The cuts 3 and 4 were actually 2 inches lower than I expected. I'm cleaning that up here too, but that may create a bit more stability in the output.

blurfl:
> The cuts 3 and 4 were actually 2 inches lower 

So when I measured center-to-center I was half a mm long! 😁

reecej:
Well cuts 3 and 4 are primarily used for chain sag correction, so some error in height there (where we see the most chain sag effects) could create some of the widely varying values. Out of curiosity, could you measure the distance of the cuts 1/2/3/4 from the top/bottom? They should be around 10, but I suspect 3/4 are closer to the bottom than that.

blurfl:
Cuts three and four end about 7 1/2 inches from the bottom of the sheet.

reecej:
Thank you, they should end about 9 1/2 inches. Correcting that with the aforementioned changes!

reecej:
Alright, hopefully I have it all sorted out now. The cuts should all be approximately 9 1/2 inches away from the edges (after the machine is calibrated). The other thing I would request to be validated is that the motorYoffset value is correct. With moving cut 5 I had to adjust the calculation for motorYoffset, so I'm more just looking to ensure that after calibration the home location is truly the center of the sheet.

blurfl:
Ok, I’ll check it.

blurfl:
The marks are 9.5"" from the upper and lower edge, and 10" in from the sides. The 3-4 measurement was 1mm less than the previous runs (this time 1929.4 ) with the other cut order. Cuts 1-2 were the same, 1930.4. The new measurement for motoroffsety was within 1mm of the previous. I measure it to be 475mm, the calc guesses 464.4.Probbly close enough 👍 The rR came out 128.0 instead of 127.7 and the chain sag came out 35.75 instead of 29.05
 I say "ship it!"

reecej:
Thanks for your testing! Sounds like your sag correction value was right in the middle of the values you were getting before, which is a good sign.

blurfl:
Here's a run with after using the most recent cal. routine.
calibration benchmark test report sheet - 2453887 -1
I notice that even though some squares differ in size, none are parallelogrammed. Most satisfying! 
 Smething funny about the right side of my machine; why not symmetrical?

reecej:
Thanks @blurfl! From the horizontal lengths, it looks like that combination of rotationRadius and chainSagCorrection is pretty good for you. Are the vertical lines all supposed to be 900?

That is also very interesting about the right side of your machine. Bar saw similar errors on the right side, but also saw errors on the widths of the 100mm squares on the left side, which you don't seem to.

blurfl:
I’m not in a place to check the file - it’s taken from Bar’s open PR, I think it’s the same one he’s been running.

BarbourSmith:
> Something funny about the right side of my machine; why not symmetrical?

My right side is funny too! Let's investigate. 

The Y-span should be 900mm so you are looking way closer than if it was 800mm 😀 

I am sure we can continue to improve, but we've already proved conclusively that this pull request brings us a roughly five fold improvement in accuracy. It's been tested with both the ring and linkage systems. It is fantastic, and I am going to merge it here. We can continue to make tweaks in further pull requests.

I want to say a huge thank you to everyone who worked on this and especially @reecej for making it all happen. This is huge. Totally massive. This is open source at it's best with people working together to build something. Great work everyone! 👍 👍 🍾 🎆

reecej:
Thank you everyone for your assistance with this process. I greatly appreciate your input of time and testing to get this sorted out. It was a great team effort!

davidelang:
@BarbourSmith did you ever try attaching a 2x4 across the top of your machine to 
see what difference it makes to the accuracy?

BarbourSmith:
I haven't yet. I started to and then I decided that enough of the play was in the plywood motor mounts that I really think the right solution is to attach the motor brackets directly to a beam like you show in your design. That way there are the minimal number of joints which could have play in them.

PR #559, opened 2018-01-11
Make move here/mark here reference machine units, not gcode units

### LINKED ISSUE #549: MoveHere/MarkHere conversion error in Inches mode
Platform: OSX, GC v1.02

The MoveHere and MarkHere functions fail to convert from mm to inches when GC is in Inches mode. This can cause the MarkHere mark to be out of the frame in GC, and the sled to try to leave the workarea to reach the MoveHere point. Looking at the values calculated in modernMachine.py, there doesn't appear to be a conversion applied. The units parameter in data.py doesn't get updated by frontPage.py; is there an easy way to ask frontPage.py for the appropriate conversion?

Steps to replicate the issue:
Start from 0,0 im MM mode. 
Place a mark at x=10mm, y=0mm.
Use MoveHere to traverse correctly to this mark.
Re-center to 0,0.
Change to Inches mode.
Use MoveHere to traverse with a target near the 10mm,0mm mark - beware, the sled will move beyond the target to x=10 inches

Printing what look like the pertinent values from modernMenu.py from a 10mm move in MM mode:
touch.pos[0] 982.0
touch.pos[0] 646.0
self.xPos 9.74437335612
self.yPos 0.28295272448
self.parent.pos[0] 883.183236984
self.parent.pos[1]643.130603344
self.parent.scale 10.1409048488
Sending: G0 X9.744 Y0.283  

Printing what look like the pertinent values from modernMenu.py from the same move in Inches mode:
touch.pos[0] 980.0
touch.pos[0] 644.0
self.xPos 9.54715229648
self.yPos 0.0857316648395
self.parent.pos[0] 883.183236984
self.parent.pos[1]643.130603344
self.parent.scale 10.1409048488
Sending: G0 X9.547 Y0.086

BarbourSmith:
The move here and mark here commands operated on the scale factor of the gcode file. They should be using the same units as the machine becuase they can send information to the machine.

This PR fixes #549

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -355,8 +355,13 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
         Move the machine to a point selected on the screen
         
         '''
-        xTarget = '%.3f'%(xPosition/self.canvasScaleFactor)
-        yTarget = '%.3f'%(yPosition/self.canvasScaleFactor)
+        if self.data.units == 'MM':
+            scaleFactor = 1
+        else:
+            scaleFactor = 25.4
+        
+        xTarget = '%.3f'%(xPosition/scaleFactor)
+        yTarget = '%.3f'%(yPosition/scaleFactor)
         commandString = 'G0 X' + str(xTarget) + ' Y' + str(yTarget) + " "
         
         self.data.gcode_queue.put(commandString)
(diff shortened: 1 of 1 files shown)


blurfl:
Nicely done, thanks for addressing it so quickly 😊

BarbourSmith:
It's easy when you give me such clear instructions for how to replicate the issue 😊

PR #562, opened 2018-01-11
Remove the logic that prevented the z-axis prompt

blurfl:
'Return-to-Center' and 'Home' had logic that prevented the Z-axis prompt from appearing.


BarbourSmith:
That was quick! Fantastic work! 👍 👍

Issue #382, opened 2017-07-31
Some on-screen type _very_ small


blurfl:
Not certain when this cropped up, but it was correct in release 0.82.
<img width="800" alt="screen shot 2017-07-31 at 4 23 36 pm" src="https://user-images.githubusercontent.com/1851315/28802771-105f0970-760d-11e7-9dfa-9fb015af586d.png">

BarbourSmith:
Oh, I'm glad you pointed that out. I've been messing with exactly that today 😬

blurfl:
Furthermore, the top two error sliders draw part of their text ouside the box - only 'Error' shows for each.

davidelang:
what sort of system are you using? what screen resolution?

blurfl:
The system is OS X. The labels drawn with screen resolution 1024x640; anything large shows the issue.

blurfl:
I’ve played around and made some progress. I’ll post a commentwith details later, after I get four more post holes dug 🙂

blurfl:
I swear I had both Windows and OSX displaying the Simulation control labels correctly last night. Tonight Windows is displaying wrong. I'm unable to get these things pounded into github for a day or so, so some observations:

I did find that Window.size[0] is width, [1] is height. Printing window.size on my OSX system shows the retina display size:
window w, h
(5120, 2706)

The tiny text is fixed by font_size: '10sp' as you suggested

On OSX I was having luck formatting the simulation labels so:
            Label:
                text: "Motor Spacing\nError: 0mm"
                text_size: self.width, sp(self.height)
                font_size: (self.height - 6)/2
                halign: 'left'
                valign: 'middle'
                id: motorSpacingErrorLabel

Sorry for the disjointed message, I hope something in it is of use>

BarbourSmith:
Fixed in #565! Way to go @blurfl 👍 👍

Issue #445, opened 2017-11-03
Text on buttons and in windows shows up strangely on lower resolution screens


BarbourSmith:
The button text will over run the edges of the buttons, as will the text in popups as shown in the picture below:

![image](https://user-images.githubusercontent.com/9359447/32389408-c680bfc2-c087-11e7-989e-eda7ac1876b0.png)

This could be fixed by either adding a scroll bar to make all of the text visible, or by making sure the text resizes to fit the window.

PR #565, opened 2018-01-12
Fix-Simulation-Slider-Label-and-machineLabel-Font-Size for large displays

blurfl:
machineLable labels and sliders used in Simulation carousel should use sp() for font_size to accomodate large displays. Change the divisor used to calculate the font_size in the Sliders' labels from 2 to 3 to get the text to fit within the labels on larger screens (didn't figure out how to do this using sp() instead of changing the diviosor...)

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -695,7 +695,7 @@
                 size_hint_x: .4
                 Label:
                     text: "Left Sprocket"
-                    font_size: 20
+                    font_size: '20sp'
                 Label:
                 Label:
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Fantastic! Does this fix issue #382 ?

Could issue #445 be addressed the same way?

blurfl:
> Does this fix issue #382 ?

Yes, that is the 'machineLabel' change.

> Could issue #445 be addressed the same way?

I'll look at it.

PR #567, opened 2018-01-13
Use Scale-independent Pixels in more parts of Simulation carousel

blurfl:
Scale-independent Pixels helps solve the text size issues in the labels. Haven't found a Linus system to test on yet...

Screenshots after fix:
OSX retina
simulation patch osx retina

WIN10 1368 x 768
simulation patch win136x768

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -1190,11 +1190,11 @@
         Label:
             text: "Not yet computed"
             id: machineLabel1
-            font_size: 10
+            font_size: sp(10)
         Label:
             text: "Not yet computed"
             id: machineLabel2
-            font_size: 10
+            font_size: sp(10)
     GridLayout:
         cols: 2
         size_hint_y: .9
@@ -1214,7 +1214,7 @@
             Label:
                 text: "Motor Spacing\nError: 0mm"
                 text_size: self.width, self.height
-                font_size: (self.height - 6)/2
+                font_size: sp(self.height)/sp(3)
                 id: motorSpacingErrorLabel
             Slider:
                 min: -50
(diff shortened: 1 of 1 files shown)


BarbourSmith:
This looks beautiful! 

I'm seeing some conflicts (probably because this branch was created before your last PR was merged). 

image

In all of these cases where it's asking me to pick one version or the other, I just want to pick the one which uses the sp, right?

blurfl:
the 'sp' one, right. Sorry for the confusion!

davidelang:
hmm, given the suggestion from bar to create a way of quantifying the accuracy, 
I wonder how hard it would be to add this to the simulation.

bonus points for anyone who manages to add 'frame flex' (aka errors that scale 
with tension) to the simulator.

:-)

blurfl:
Here's a screenshot from a Ubuntu Linux box at 1920x1200:
simulation patch linux 1920x1200

Trying Linux at 800x600 is a mess, but then so is the calibration carousel 😟

blurfl:
I've done some more work in the Simulator - added "Bar's-score", though I don' think I'm calculating correctly:
Average the long-horizontal-cut error then average the long-vertical-cut error and average those two for the first digit. 
Average the box-horizontal-cut error then average the box-vertical-cut error and average those two for the second digit.
Is that the idea?

BarbourSmith:
Wow! That was fast!

I think that sounds right, but I'm not 100% sure I understand so I'm going to repeat it.

Of a score like `5.43-2.10` the 5.43 is the average of the absolute value of the error in all of the long cuts (the 900mm cuts and the 1905mm cuts).  The second number the 2.10 is the average of the absolute value of the error in the smaller 100mm square cuts. 

For example if the long measurements were 899, 901, 900, 1905, 1905, 1906 that would give us absolute values of the errors of 1,1,0,0,0,1 which would have an average of 0.50

blurfl:
Thanks for the clarification! What do you want to call the figure?

BarbourSmith:
Oh, I don't know. We don't have a name yet. What about "Calibration Benchmark Test" or is that too boring?

blurfl:
How about "Calibration Benchmark Score"?

BarbourSmith:
That sounds excellent!

PR #570, opened 2018-01-13
Simulation-use-current-kinematics-setting

blurfl:
Start the simulation using the kinematics type found in groundcontrol.ini. Allow the use of the other type for simulation but do not change the .ini file setting.


BarbourSmith:
This is a fantastic idea! These things that just make the user experience a little better really add up to make the program more enjoyable to use 👍 👍

PR #574, opened 2018-01-13
Simulation format results to four decimal places

blurfl:
Addresses Issue #537 - 12 decimal places not necessary.


blurfl:
Used four places, from comment by @davidelang 

> if you are mm on in, for mm, 2 digits is enough, for in, 4 would be
better

BarbourSmith:
👍 👍Excellent collaboration

Issue #571, opened 2018-01-13
Simulation calculations values - float or integer


blurfl:
Most of the constants used in the Simulator are without decimal values. Adding '.0' to them makes a noticeable difference in the results, visible in the values calculated with zero error. Should these constants be explicit floats? I'm not sure how Python2.7 handles them...

BarbourSmith:
Yes! Great catch. I think making them explicit floats is an excellent idea

blurfl:
Good-o, I'll try to keep this PR from going crossways 😉

blurfl:
Solved in pr #574

PR #575, opened 2018-01-13
convert variables to explicit floats

blurfl:
Addresses Issue #571  - improves accuracy of simulation results


BarbourSmith:
Good catch! This is a subtle one, out of curiosity how did you notice what was going on?

blurfl:
> how did you notice what was going on

The "Calibration Benchmark Score" calculation gave the wrong result whedn fed the example values you provided...

PR #576, opened 2018-01-13
Create calibration benchmark score in simulation

blurfl:
Add the result text to the machineLabel1 string containing the error texts for the triangular calibration and the 800mm centered square.

The formula is described:
"The benchmark is two numbers like 1.23-3.45 where the first number is the average of the absolute values of the error in the long measurements and the second number is the average of the absolute values of the error in the short (100mm) measurements. The first number tells us about big distortion across the sheet, the second number tells us about localized distortions. A perfect test would give 0.00-0.00.

In a score like 5.43-2.10 the 5.43 is the average of the absolute value of the error in all of the long cuts (the 900mm cuts and the 1905mm cuts). The second number the 2.10 is the average of the absolute value of the error in the smaller 100mm square cuts.

For example if the long measurements were 899, 901, 900, 1905, 1905, 1906 that would give us absolute values of the errors of 1,1,0,0,0,1 which would have an average of 0.50"


BarbourSmith:
This is fantastic! Here is what it looks like:

image

It's fun to change the settings and get immediate feedback about how the score changes. What a great idea.

This PR is showing conflicts with the last PR and I'm not quite sure which lines to keep and which can go, @blurfl do you feel good about resolving the conflicted lines?

BarbourSmith:
I think that the auto merge feature in Git is giving us trouble with this one. I resolved the conflicts it was detecting, but I'm still seeing some things that don't look right when I check the files changed, for example it wants to go back to not using floats which seems wrong:

image

It might be easier to just open a new pull request 😕

blurfl:
> It might be easier to just open a new pull request

Ok, I think you're saying after you get the other PR's merged, I should make a new PR with the floating point fix. I'll be glad to 😊

blurfl:
Wait, did I misunderstand which PR you want? You want this one, don't you... Sorry, half-a-cup low on the coffee 😊

BarbourSmith:
It seems like this pull request is overwriting your earlier changes so I vote we close this one without merging it, and make a new one that adds the benchmark score to the simulation

blurfl:
Now I understand, go ahead and close it, I've got a re-do of this ready to send.

PR #579, opened 2018-01-16
Adds a button under "advanced" to load the callibration test file

BarbourSmith:
I felt like if we want to make it easy to run the test file it would be a good idea to include it in GC so that anyone can easily run the same test


BarbourSmith:
If anyone would be willing to test on an OS other than Windows I would be much obliged, anything which has a hard coded path could have issues across platforms

blurfl:
This works under OSX and Ubuntu 17.10 - nice!

How about a button to calculate the "Calibration Measurement Score" from my results? 😁

BarbourSmith:
Great suggestion @blurfl! On the to-do list!

PR #580, opened 2018-01-16
Create calibration benchmark score in sumulation redo

blurfl:
Add the result text to the machineLabel1 string containing the error texts for the triangular calibration and the 800mm centered square.

The formula is described:
"The benchmark is two numbers like 1.23-3.45 where the first number is the average of the absolute values of the error in the long measurements and the second number is the average of the absolute values of the error in the short (100mm) measurements. The first number tells us about big distortion across the sheet, the second number tells us about localized distortions. A perfect test would give 0.00-0.00.

In a score like 5.43-2.10 the 5.43 is the average of the absolute value of the error in all of the long cuts (the 900mm cuts and the 1905mm cuts). The second number the 2.10 is the average of the absolute value of the error in the smaller 100mm square cuts.

For example if the long measurements were 899, 901, 900, 1905, 1905, 1906 that would give us absolute values of the errors of 1,1,0,0,0,1 which would have an average of 0.50"


blurfl:
Wait!! I've got some cleanup to do!

blurfl:
Ok, that fixed the issue, ready for prime time. Sorry for the delay - he who hesitates is sometimes  saved...

BarbourSmith:
Perfect! All is well in the world! 👍 👍 

image

PR #582, opened 2018-01-16
Properly Treat Error Responses as Equivalent to ok

krkeegan:
This is a small fix.  The Grbl spec will send error: xxxx responses with no ok response.  This tells GC to properly move on to the next message when such responses are received.


BarbourSmith:
Thank you for doing a really good job of reading the standards and making sure we stick to them. Those sorts of things may be small fixes when taken independently, but together they make up the difference between a project which is a pleasure to use and work on and one that is dificult 👍 👍

PR #585, opened 2018-01-19
Restore lost octothorpe

blurfl:
Lost the leading octothorpe for a comment line.


krkeegan:
Ha, learned a new word.

blurfl:
I've confirmed that buildozer is satisfied with the file after this correction is made.

PR #587, opened 2018-01-20
Fix Issue 586

reecej:
This fixes [issue 586](https://github.com/MaslowCNC/GroundControl/issues/586) by allowing measurement values to be re-entered if a successful calibration was not able to be performed.


blurfl:
This works a treat! Thank you, a very good improvement 💯👍

Issue #520, opened 2017-12-27
If No groundcontrol.ini file is found, user gets warning about unable to locate gcode file


krkeegan:
Try deleting or renaming groundcontrol.ini.

Then open GC.  You will get a warning about unable to load gcode file.

We should probably capture this somehow as all new users will see this error first.

ladams00:
New user here.  I saw the same, didn't know what to think and ignored the error.

PR #590, opened 2018-01-23
Added a button and code to clear gcode.

thormj:
I wanted to be able to clear the g-code buffer to clean up the screen.
Plus, this is a way for me to see how to use GitHub...

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -156,8 +156,7 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
                     else:
                         self.data.zMoves.append(index)
         except:
-            if filename is not "":
-                self.data.message_queue.put("Message: Cannot reopen gcode file. It may have been moved or deleted. To locate it or open a different file use Actions > Open G-code")
+            self.data.message_queue.put("Message: Cannot reopen gcode file. It may have been moved or deleted. To locate it or open a different file use Actions > Open G-code")
             self.data.gcodeFile = ""
     
     def centerCanvas(self, *args):
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -110,6 +110,10 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
         '''
         
         filename = self.data.gcodeFile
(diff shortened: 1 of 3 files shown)


blurfl:
This does what it says on the box. It leaves the gcode file selection in groundcontrol.ini alone so that on next launch the previously selected gcode file is back and the location is preserved. 
 You've mastered the GitHub thing on the first go!
 I probably won't use this option very often, though. I've got an empty gcode file with only the setup lines (G90, G20) that I use as a starting point, it accomplishes much the same thing.

BarbourSmith:
Fantastic! Excellent mastery of GitHub!

For anyone following along here's what the button looks like:

image

How do we feel about moving the button over into one of the other less crowded columns? Maybe replacing that "zero z" button that has been moved into the z-axis popup?

To make changes, just make a new commit to the same branch as you did before (ClearGCode) and it will show up here automatically 👍 👍

thormj:
I thought of that, but I kinda like how each column is dedicated to something "View" (everything dealing with code) (maybe should be "Code" instead),  Run (if you hit these buttons something happens).

OTOH, I suppose you could put a button on the file picker to "clear".

I was also trying to pass in python main.py -c:data:GCodeFilename:"" to see if it would blank a file, but evidently I don't understand command line parameters much...

blurfl:
I've had success with a command line argument to groundcontrol in the past, but can't find the code right now.
Take a look at this [guy's example](https://gist.github.com/jegger/28d3e295ae690e2fa2ac) to passing a command line value into kivy. I started from there, I think.

UIElements/gcodeCanvas.py
```diff
@@ -156,8 +156,7 @@ def reloadGcode(self, *args):
                     else:
                         self.data.zMoves.append(index)
         except:
-            if filename is not "":
```
BarbourSmith: I think we added this to fix issue #520 . Sorry to nitpick!

blurfl:
That patch for issue #520  isn't needed with the code brought by this PR. This handles the missing file by clearing the gcode list.

BarbourSmith:
Great catch!

BarbourSmith:
Sorry for missing that

BarbourSmith:
In that case it looks great to me! 👍 👍

Issue #592, opened 2018-01-27
Display velocity for Maslow and current g-code line on the screen


thormj:
Figuring between Z and not connected, I'd like an indicator that says:
Vel: 30 ipm  | 30 ipm

Where the first number is from Maslow, and the 2nd number is from the current line of g-code.

thormj:
Is there a (preferably auto-generated) list of g-code commands somewhere?

thormj:
I'll take this one and run with it for a while...

BarbourSmith:
I believe that there is a list of supported commands on the wiki page, but it is not automatically generated (which would be slick)

We've been talking about setting up circle CI integration to automate tasks like that which I am happy to help with on my end

BarbourSmith:
I think you nailed this one, good to close?

thormj:
Yep.

PR #599, opened 2018-01-30
Add vel readout

thormj:
Fulfills issue #592 
Needs corresponding firmware update

### DIFF
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -252,34 +252,42 @@ class FrontPage(Screen, MakesmithInitFuncs):
     def upLeft(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G00 X" + str(-1*self.stepsizeval) + " Y" + str(self.stepsizeval) + " G90 ")
+        self.gcodeVel = "[MAN]"
         
     def upRight(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G00 X" + str(self.stepsizeval) + " Y" + str(self.stepsizeval) + " G90 ")
+        self.gcodeVel = "[MAN]"
 
     def up(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G00 Y" + str(self.stepsizeval) + " G90 ")
+        self.gcodeVel = "[MAN]"
 
     def left(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G0 X" + str(-1*self.stepsizeval) + " G90 ")
+        self.gcodeVel = "[MAN]"
         
     def right(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G0 X" + str(self.stepsizeval) + " G90 ")
+        self.gcodeVel = "[MAN]"
         
     def downLeft(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G00 X" + str(-1*self.stepsizeval) + " Y" + str(-1*self.stepsizeval) + " G90 ")
+        self.gcodeVel = "[MAN]"
 
     def down(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G00 Y" + str(-1*self.stepsizeval) + " G90 ") 
+        self.gcodeVel = "[MAN]"
 
     def downRight(self):
         self.jmpsize()
         self.data.gcode_queue.put("G91 G00 X" + str(self.stepsizeval) + " Y" + str(-1*self.stepsizeval) + " G90 ")
+        self.gcodeVel = "[MAN]"
     
     def zAxisPopup(self):
         self.popupContent      = ZAxisPopupContent(done=self.dismissZAxisPopup)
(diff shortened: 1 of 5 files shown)


blurfl:
I've loaded the firmware that goes with this, and tried without success to run a small cut file.
 I see this error (when run from command line).
> ValueError: FrontPage.gcodeVel accept only str

thormj:
Aye... I guess that's what I get for checking it without the machine (I checked the firmware bit, but didn't try to actually run after adding the g-code parser).  Should be fixed now (I formatted it into a string).  If you want to wait 'til midnight or so I'll double check it on my machine and confirm it works (or fix it and let you know when a confirmed working version is pushed).

blurfl:
I may be getting ahead of things, but IK tried the change and got this error:

>    File "/Users/xxxxxx/Desktop/GroundControl-AddVelReadout/UIElements/frontPage.py", line 163, in onIndexMove
     self.gcodeVel = "{0:.2f}".format(F)   #Otherwise, it stays what it was...
 ValueError: Unknown format code 'f' for object of type 'str'

thormj:
Ok... I'll play with it tonight and get it beaten into shape.

BarbourSmith:
The code looks nice and clean 👍 

Let me know when to test

blurfl:
Perhaps inelegant, but changing the line in frontPage.py to 
>self.gcodeVel = "{0:.2f}".format(float(F.group(2)))   #Otherwise, it stays what it was...
runs without error and reports the value of the F parameter on the gcode line. 😊
I think the re.search was returning an array?

thormj:
Doh!  I forgot the next line (gCodeCanvas.py:302) which turned the RE.SearchObject back into a string that then got converted to a float.  I just left it as a string here so it should match the g-code.

blurfl:
I think the patch should use item[1] of the list... item[0] is "F"... Just ran it to confirm 🙂

blurfl:
There's a typo on line 113 of frontPage.py, RedoutVel instead of ReadoutVel . This keeps the value from appearing on the screen. 
 Once the value is showing, I see that the left-hand number in Vel. is the current (actual, measured?) velocity of the gcode line being cut, while the right-hand number is the feedrate of the _next_ line that will be executed. Kind of neat to watch the velocity change, but maybe there's a way to display the feedrate of the line currently being executed instead of the next one...

thormj:
Ok, changed.  We should be good to go.  Sorry about the immature patch.

blurfl:
I hope I'm not being pushy! 😀 I'm very much enjoying the instrumentation you're providing!

thormj:
Nah, we're good... I was apologizing for submitting a broken pull request.  @BarbourSmith  we're good to go on both pieces.  It won't do much now, but when the firmware gets reset we should be cooking with gas.

blurfl:
The reference to the previous line works correctly to sync the two velocity displays.
But...
I feel badly about this - with the revised code, the array reference to member [0] is needed instead of [1]. I re-loaded the version I used when I made my comment, and [1] was what worked here at that time. I feel I've just made more work for you than if I'd just backed off and waited. I was so excited to see the speedometer and get it going...

blurfl:
I'm hoping this can get merged in time for the release tomorrow -a  nice addition to the instrumentation !

blurfl:
This works great! 👍💯

thormj:
There's been a discussion of "Why can't we do it all from GroundControl instead of asking the firmware"... (Doh!  Duh!)
6922fe2 implements that (I'm away from my Maslow at the moment), so I don't really know how well it will work... @blurfl, @BarbourSmith if this works... we can ignore the firmware bit...

blurfl:
The latest commit changed frontPage.py has caused setPosReadout() to fail resulting in this message:
>One Machine Position Report Command Misread

It looks to me like there are some issues
 - xpos, ypos and zpos need to be changed to xPos, yPos and zPos
 - don't lastpos and lasttime need to be referred as self.lastpos and self.lasttime?
 - python doesn't seem to like to take the sqrt() of the object to the left of the = sign 
 - the x-y-z calculation can be negative, don't we need to take the root of the abs() of that?

With those changes, I got setPosReadout() to stop causing the 'Misread' message, but the Val values are not right. Does the calculation for Vel need some parentheses?

screen shot 2018-01-31 at 6 45 01 pm

This is a good step in the right direction. It would make the change to the firmware unnecessary, since we calculate Vel instead of read it from the firmware.

thormj:
Yeah, I boogered that in so many ways. /Me hangs head in shame.
I've fixed it, slowed it down (every time through was too noisy), removed all the comm bits and tested it.
801e6e9 *is* a good one.

Should I close the other pull request as not needed or is that something the maintainers do?

blurfl:
I think you can close it yourself, and make a comment about with this PR# in it to explain the closing.

blurfl:
This works a treat! I'm having a great time sending my sled all over the place watching the speedo! (Not cutting, though...) Won't be long before folks will want to 'tweak the governor' and get more speed! 💯👍😁

BarbourSmith:
Oh man, this IS fun! 

Great contribution @thormj !!! 👍 👍 💯 

I'm going to have fun keeping an eye on those numbers

PR #600, opened 2018-01-31
Change default rotational radius

madgrizzle:
Changed the default rotation radius used in triangular kinematics to 250 from 100.  I don't know of any linkage kit that uses a rotation radius of 100 and its caused a few headaches being so low.


BarbourSmith:
Excellent plan! 

The ring kit is around 140 so we might run into issues down the road, but hopefully we'll have it added to the calibration process by then

Issue #595, opened 2018-01-28
After Z cal, move to safe height


thormj:
Since it was touching the board, I was wondering if it was going to lift up before moving to the cutting position.  When you leave the Z-cal step (if using motors), raise the router to the "safe height"

BarbourSmith:
Good point! It would be easy to send a lift command right after that step is completed

BarbourSmith:
Is this guy good to close?

blurfl:
I would say, yes

PR #601, opened 2018-02-01
Zero z 595

thormj:
Fixes Issue  #595  and adds "ZeroZ Touch" and e-stop to the Z-calibration carousel

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -110,6 +110,9 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
         '''
         
         filename = self.data.gcodeFile
+        if filename is "":
+            self.data.gcode = ""
+            return
         try:
             filterfile = open(filename, 'r')
             rawfilters = filterfile.read()
--- a/UIElements/measureMachinePopup.py
+++ b/UIElements/measureMachinePopup.py
@@ -1,6 +1,6 @@
 '''
 
-This allows the user interact with the z-axis when it is the content of a popup
+This is the big calibration procedure
 
 '''
 from   kivy.uix.gridlayout                       import   GridLayout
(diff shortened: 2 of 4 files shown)


blurfl:
Although I use the touch plate frequently, I don't think it belongs in the standard cal sequence. It requires extra equipment (which is shamefully poorly documented 😳) and isn't widely used. 
My approach is to define one of my macros for this purpose (wish there were more macros available...) I also have a single-purpose .nc file to load which does only this thing. 
Now that the workarea accuracy is better near the edges, I've been thinking of putting a grounded screw head flush way up in a corner, out of the way of cuts. Still need to attach the clip to the bit, though.

If others want to persue this, it will need some very good pictures and description on the page.

BarbourSmith:
I agree with your concerns @blurfl, I also think that @thormj is onto something important which is that we need to make it more clear to people that the z-axis touch probe is an option.

I think we have a little bit of a chicken and the egg problem where nobody uses it, because the feature isn't that well documented, and it's not that well documented because nobody uses it.

How would we feel about adding it to this z-axis popup instead:

image

I am going to tidy up the calibration process very soon (as soon as the frame is all squared 😉 away), and one of the changes I want to make is to use the regular z-axis popup in that step where you move the z-axis up and down.

thormj:
I might've been possessed by creeping featurism when I added the "z-touch"... for people with a flush touch area, they'll probably need to move to the spot and then bring up the Z menu, and it isn't really needed in the machine setup dialogs (except for "Do I have a powered Z"), though I do like the idea of "if you make it through this, you'll know your machine is setup properly and is functional"

thormj:
Ooh... Eye have an apostrophe - how about we change this so the screen has an "enable z-touch" switch on it that puts/removes a z-touch button on the popup menu?  Probably replace the picture  with 2 pics (one of powered z and one that shows the wiring/pic of the z-touch feature, and add some text saying "if you have a z-touch, you probably want to move to the correct position and the use the Z button instead of calibrating it here"

blurfl:
> tidy up the calibration process very soon

While you're in there how about arranging the separate 'Automatic Chain Cal...' screen to allow exiting when the sprockets are aligned? Currently there is no clean way out of 'Automatic Chain Cal...'. That or make the sprocket alignment available to the manual chain cal screen. I think that if the chain management arrangement goes horizontal the manual cal will become more important.

BarbourSmith:
@thormj See what I've done in #604 

How does it fit with what you are saying?

thormj:
Taking @BarbourSmith's commit and running with it instead

Issue #597, opened 2018-01-28
Add an Abort to calibration sequence


thormj:
There's no obvious/easy way to abort calibration (eg, I did 90%, but then realized I forgot to hook up my Z axis... and just plugging it in didn't work).  So I didn't want to overwrite the values...  Not sure where it saves the info to the machine, but there's no way to get out without going to the end and saying "close."

blurfl:
I agree you should be able to back out (or skip forward, I think) without anything getting changed unless you use the 'Continue' button (rename it 'Accept Values'?). Presently, if any values have changed in earlier steps, GroundControl.ini gets re-written when the step is entered. Bug or Enhancement? 

Currently the z-axis step comes after the 'review and accept' step. I think that was by design 🙂. Changes to the 'z-axis present' switch in that step seem to work as expected. I unhooked my z-axis and went to that step and clicked a movement button to try to replicate the error. After I plugged the motor back in I was able to use the movement buttons as expected. If you have a way to replicate your error situation, that sounds like a good candidate for a new issue.

thormj:
I'll double check if it's truly an error and will create a new issue for "plugging z in halfway through calibration"...

I'd still like a "Panic-Quit-Abort" button for if you had done a calibration previously, but you b0rked something this time around (like trying to extend the chain 3.8m ...)

BarbourSmith:
Should be good on this one!

PR #602, opened 2018-02-01
Calibration abort

thormj:
Implements Issue #597.

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -110,6 +110,9 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
         '''
         
         filename = self.data.gcodeFile
+        if filename is "":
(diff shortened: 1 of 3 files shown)


BarbourSmith:
This will add a close button (which I think is excellent), but it should be worth noting that leaving the calibration process midway through can have unexpected consequences like having the chains not extended. You would think that would be obvious, but we do get a lot of questions when we let the machine get into weird states

BarbourSmith:
Could we trigger a popup when the close button is pressed saying that the machine may be left in an unusual state? We can trigger a popup easily by just putting a text string on the dara.message queue which starts with the letters "Message: The machine may hav...."

BarbourSmith:
This looks good to me! 👍 👍 

If we decide that a message is in order to warn that the calibration process was exited part way through let's make that change in a follow up PR because I'm going to merge this one!

Issue #354, opened 2017-07-20
Calibration Machine Dimensions Step 7


jbnimble:
Use the "main" Z-Axis movement buttons to extend and retract the bit, and take advantage of the ability to set how far the bit is moved per button press. This should help familiarize new users on how the main buttons and size per button click actions work.

BarbourSmith:
Still a great idea

Issue #349, opened 2017-07-20
Calibration Machine Dimensions Step 1


jbnimble:
Use the "main" movement buttons (Left/Right in this case) to extend and retract the chain, and take advantage of the ability to set how far the chain is moved per button press. This should help familiarize new users on how the main buttons and size per button click actions work.

BarbourSmith:
This is still a great suggestion that I think we should add

blurfl:
So this would be an enhancement, yes? Not required, but nice to have...

PR #604, opened 2018-02-01
Use normal z-axis popup during calibration

BarbourSmith:
This PR starts to address the feature request in #349 which asks to use the normal machine controls as much as possible during the calibration process. 

The "Adjust the z-axis to depth zero" step will now use the same controls as the regular interface.

Here is what the step looks like:

image

And here is what the popup looks like:

image

The button to open the popup is disabled if the z-axis is not installed:

image


BarbourSmith:
Fixes #354

PR #605, opened 2018-02-02
Make it possible to extend chain by an arbitrary length

### LINKED ISSUE #350: Calibration Machine Dimensions Step 2
Use the "main" movement buttons (Left/Right in this case) to extend and retract the chain, and take advantage of the ability to set how far the chain is moved per button press. This should help familiarize new users on how the main buttons and size per button click actions work.

BarbourSmith:
Instead of requiring you to extend the chain by 1000, 100, or 10mm you can now extend it by an arbitrary length entered using the standard text entry widget:

image

image

Fixes #350 

Should not be merged until after #604 

Needs to be fully tested during a calibration process before merging

### DIFF
--- a/UIElements/adjustZCalibrationDepth.py
+++ b/UIElements/adjustZCalibrationDepth.py
@@ -1,8 +1,8 @@
 from   kivy.uix.widget                    import   Widget
 from   kivy.properties                    import   ObjectProperty
-from   kivy.properties                    import   StringProperty
 from   UIElements.touchNumberInput        import   TouchNumberInput
 from   kivy.uix.popup                     import   Popup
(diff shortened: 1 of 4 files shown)


blurfl:
I've wished for this since day one! The picture shows 'Number of Links' at the top left, 'Inches' in the dialog and Dist(MM) in the Settings panel, an opportunity for refinement...

BarbourSmith:
I fixed the "number of links" at the top, but I haven't fixed the inches/mm issue yet. I am going to merge this one anyway so that we can build off of it, and work on a fix for the inches/mm confusion...great catch!

PR #606, opened 2018-02-02
A proposed fix for #306

BarbourSmith:
This is a proposed fix for the issue where the machine will not move again after pressing the stop button.

The solution is to indicate that when a stop has been sent it will free up the entire buffer. As soon as the OK message is recieved to indicate that the stop command has been processed the entire buffer is marked as free.

One more change I would like to make if we like this solution is to change all of the hard coded 256s in there to be a #define so it is more clear why that number is used and the code can be easily modified for different buffer sizes.

Closes #306


krkeegan:
bufferSpace is already defined, we just don't seem to use it.

BarbourSmith:
I think that bufferSpace is the amount of space available in the buffer so I added a bufferSize variable which is the size of the buffer when empty and added a comment to clear it up

PR #607, opened 2018-02-02
Add Chain Wrap Direction

reecej:
This adds a configuration setting to select which way the chain wraps around the sprocket to go to the motor. Changes are also made to update the simulator and triangular kinematics calibration code, as well as add a calibration step to select the appropriate configuration.

Since I don't have my Maslow yet I was not able to take pictures showing the top versus bottom configurations in the calibration process, so I did my best to utilize already-available images. Additional images may help new users here though.

Note that this does not update the quadrilateral kinematics, so using quadrilateral kinematics with the chains on bottom of the motor sprockets could lead to accuracy errors.

This is the GroundControl PR associated with Firmware [PR 382](https://github.com/MaslowCNC/Firmware/pull/382).

### DIFF
--- a/UIElements/setSprocketsVertical.py
+++ b/UIElements/setSprocketsVertical.py
@@ -12,73 +12,109 @@ class SetSprocketsVertical(Widget):
     def LeftCW(self):
         degValue = float(self.data.config.get('Advanced Settings',"gearTeeth"))*float(self.data.config.get('Advanced Settings',"chainPitch"))/360.0;
         self.data.gcode_queue.put("G91 ")
-        self.data.gcode_queue.put("B09 L"+str(degValue)+" ")
+        if self.data.config.get('Advanced Settings', 'chainOverSprocket') == 'Top':
+            self.data.gcode_queue.put("B09 L"+str(degValue)+" ")
+        else:
+            self.data.gcode_queue.put("B09 L-"+str(degValue)+" ")
         self.data.gcode_queue.put("G90 ")
 
     def LeftCCW(self):
         degValue = float(self.data.config.get('Advanced Settings',"gearTeeth"))*float(self.data.config.get('Advanced Settings',"chainPitch"))/360.0;
         self.data.gcode_queue.put("G91 ")
-        self.data.gcode_queue.put("B09 L-"+str(degValue)+" ")
+        if self.data.config.get('Advanced Settings', 'chainOverSprocket') == 'Top':
+            self.data.gcode_queue.put("B09 L-"+str(degValue)+" ")
+        else:
+            self.data.gcode_queue.put("B09 L"+str(degValue)+" ")
         self.data.gcode_queue.put("G90 ")
 
     def RightCW(self):
         degValue = float(self.data.config.get('Advanced Settings',"gearTeeth"))*float(self.data.config.get('Advanced Settings',"chainPitch"))/360.0;
         self.data.gcode_queue.put("G91 ")
-        self.data.gcode_queue.put("B09 R-"+str(degValue)+" ")
+        if self.data.config.get('Advanced Settings', 'chainOverSprocket') == 'Top':
+            self.data.gcode_queue.put("B09 R-"+str(degValue)+" ")
+        else:
+            self.data.gcode_queue.put("B09 R"+str(degValue)+" ")
         self.data.gcode_queue.put("G90 ")
 
     def RightCCW(self):
         degValue = float(self.data.config.get('Advanced Settings',"gearTeeth"))*float(self.data.config.get('Advanced Settings',"chainPitch"))/360.0;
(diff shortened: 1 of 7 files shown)


reecej:
Note that the calibration process will need to be tested for the chain on bottom configuration. Specifically, I would ask someone to test that the motor spacing measurement technique using the chains still works. I added a piece of code to allow this to still be utilized with the chain on bottom of the left sprocket, but it required additional math.

blurfl:
I am gathering the parts to set up to test this, maybe get it tested today.

BarbourSmith:
You work FAST! I will test this right away also

davidelang:
I think we will still need to measure the chain across the top (slack chain will 
fall from the bottom)

BarbourSmith:
I played around with the idea of going from the bottom on one side to the top on the other side. It's not very clean and I wouldn't recommend it. 

image

image


I think we should do what @reecej and @davidelang are suggesting and just do it like we did before but send reverse commands if the chain is on the bottom.

BarbourSmith:
For the sake of smoothness and getting things done I say we don't mess with that calibration step at all in this PR and just merge the parts of this PR which add the setting to GC because that's a whole feature in and of itself. 

I've got #604 , #605, and #606 open which clean up how some of this works (letting the user extend any length of chain) and it might make more sense to start building off of those after they are merged.

reecej:
That sounds good to me. I'll pull out the calibration changes, but leave in the settings value and kinematics updates. I'll also put in the reverse commands for the motor spacing!

BarbourSmith:
Perfect!

blurfl:
When we do start looking into the calibration stuff, the chain cal CW and CCW need to be reversed. Too, it's not possible to hook the first link on the 'top' tooth, do we need to hook the second or third link there?

davidelang:
On Sat, 3 Feb 2018, Scott Smith wrote:

> When we do start looking into the calibration stuff, the chain cal CW and CCW need to be reversed. Too, it's not possible to hook the first link on the 'top' tooth, do we need to hook the second or third link there?

we should be hooking the chain to the top tooth and let it feed out to the 
outside and around the bottom and then attach things to the sled.

blurfl:
>we should be hooking the chain to the top tooth
With the chain feeding to the sprocket horizontally, the top tooth isn't enough to hold the chain in place. Give it a try. This is the same issue that was addressed in the motor distance measurement by using the third link on the top tooth. We need to change the math, and update the instructions.

reecej:
Made some updates. I removed the chainOverSprocket from the calibration process. Also configured a few calibration steps to account for the chain on bottom configuration, including the sprocket zero step, the measure motor spacing step, and the measure Y offset step. Would you guys be able to try it out?

There are a few things still outstanding. First, at some point if they want the chains on bottom we need to include that as a step, especially for the left motor as they will have been using that with the chain on top for the calibration process. Should we have them put the third link of the chain on the sprocket in that case, to prevent it falling off?

BarbourSmith:
I will test this right away! Excellent!

UIElements/setSprocketsVertical.py
```diff
@@ -12,73 +12,109 @@ class SetSprocketsVertical(Widget):
...
+            self.data.gcode_queue.put("B09 L-"+str(degValue)+" ")
+        else:
+            self.data.gcode_queue.put("B09 L"+str(degValue)+" ")
         self.data.gcode_queue.put("G90 ")
 
     def RightCW(self):
         degValue = float(self.data.config.get('Advanced Settings',"gearTeeth"))*float(self.data.config.get('Advanced Settings',"chainPitch"))/360.0;
         self.data.gcode_queue.put("G91 ")
-        self.data.gcode_queue.put("B09 R-"+str(degValue)+" ")
+        if self.data.config.get('Advanced Settings', 'chainOverSprocket') == 'Top':
```
BarbourSmith: This is really beautifully done. Its simple, but I like that it just works based on the settings and doesn't ask the user anything

BarbourSmith:
This works great for me! I'm going to merge it and then start thinking about what is next for the calibration process 👍 👍 💯

blurfl:
The change of Top/Bottom doesn't take effect until GC is restarted, is that right? If so, maybe making the change should pop up a dialog reminding the user of this, counseling to quit.

reecej:
This is where it gets a bit complicated. When the setting is changed in GC, it takes immediate effect. So right away the calibration step functions, such as the buttons to rotate the motors CW/CCW will be reversed. However, the firmware only checks the value at system startup. So really, when the setting is changed in GC the board should be restarted right away.

I think it would be good to include some disclaimer there. I can add that as a separate PR. Also, when this begins undergoing more thorough testing, just be aware that there may still be a few random bugs!

PR #608, opened 2018-02-06
Sync Settings; Store Settings in Single Location

krkeegan:
This makes a few changes to the settings design that should make our life easier.

1. This condenses all settings and their attributes into a single dict.  (no more changing default values in one place and the settings in another)
2. This changes the push of settings from all settings, to only those that are different that what is on the machine
3. Each setting is synced immediately upon change, no more need to call push settings, any change whether programatically or in the settings panel will cause a sync of that setting

This requires a simultaneous update to the Firmware, so the version number should be changed.


krkeegan:
Matching commit from firmware is https://github.com/MaslowCNC/Firmware/pull/384

blurfl:
This crashes for me (using the matching firmware) with this output:

>    File "main.py", line 274, in runPeriodically
     self.receivedSetting(message)
   File "main.py", line 236, in receivedSetting
     value, position = self.parseFloat(message, position)
   File "main.py", line 248, in parseFloat
     return (float(match.group(0)), match.end(0))
 AttributeError: 'NoneType' object has no attribute 'group'

PR #609, opened 2018-02-06
Add chain sag adjustment to simulator

reecej:
I realized the simulator did not allow you to see the effects of changing chain sag, so this adds that functionality.


blurfl:
Very nice!

krkeegan:
I think this can be safely merged.

Issue #591, opened 2018-01-25
Load/Align a "background image" on the canvas


thormj:
Enhancement: Be able to load a "background image" onto the g-code canvas and align it so you can see what the cut looks like on a picture of stock in the machine.

I've gotten a process to work (using a T3i camera, some reflectors, and OpenCV
But I need some guidance on how the menu/buttons work (thinking about adding a button to the right of the Settings button called "Background"... we can move/integrate it better later, but...)

So can someone point me to the direction of how the "Actions" button calls up the "Actions" popup that has the other menus tacked together?

Raw image:
![t3i2](https://user-images.githubusercontent.com/1596677/35373936-c03fa48e-016e-11e8-9ede-f25c59667b7f.JPG)

Aligned image:
![sq](https://user-images.githubusercontent.com/1596677/35373974-eaac7832-016e-11e8-9743-3d8b4583b734.jpg)

On GroundControl:
![image](https://user-images.githubusercontent.com/1596677/35374131-94c47c66-016f-11e8-8a5d-21a8a17a5ef1.png)

To do this, I need to know:
  Positions of 4 markers on the machine (TL,TR,BL, BR) - ok if outside workspace (I put mine 5cm outside the 4x8 area)
  Colors of said markers in OpenCV's HSV space (I'll make a picker)
  Contrast adjustment that should be applied to make it so you can see your cut lines
  Where to load the image from?

thormj:
I'll have more notes as I go along, but duplicating a color was a mistake (not mathematically, but half the time the corners come in 1234 and the other half 1243, so... more fixes).

reecej:
This is great! I really like the idea and execution.

Buttons are defined in the Kivy code: [groundcontrol.kv](https://github.com/MaslowCNC/GroundControl/blob/master/groundcontrol.kv)

Look around line 274 and you'll start to see it.

thormj:
I see the menu being declared there, but the part I'm not getting is how it goes from
  groundcontrol.kv:40 [root.show_actions] to
  [Whatever the container is for the buttons on that screen] 

I see all the menus being declared (and I even added a button to that one), but I don't see [what it is that's holding those menus together] or where root_show_actions() is located (then again, I've been having an argument with grep today so maybe it'll be better in the morning).

[...]

PR #610, opened 2018-02-06
Let's not change g-code while we're sending it...

thormj:
Short patch... don't let the user reload/load g-code while we're sending it.
Machine goes quite nuts.

### DIFF
--- a/groundcontrol.kv
+++ b/groundcontrol.kv
@@ -280,9 +280,11 @@
     Button:
         text: 'Open File'
         on_press: root.openFile()
+		disabled: app.data.uploadFlag == 1
     Button:
         text: 'Update Gcode'
         on_release: root.reloadGcode()
+		disabled: app.data.uploadFlag == 1
     Button:
         text: 'View Gcode'
         disabled: False


thormj:
And this indicates a further problem, but I'm not sure how to explain it:
If you're running "Corner-1.gc"
And you load in "Dragon.gc"
The red crosshairs stay on the path of Corner-1 until the cut is done.
The black crosshairs "follow" Dragon.
When the cut is done, the machine starts slewing towards where Dragon is, but the cursor on Dragon doesn't stop... it keeps moving.  If you press pause, the black crosshairs stop and the machine slewing stops,... and there's no way to tell the machine to "align yourself" except if the "slewing" is faster than the "cutting" on Dragon1.  And by slewing, I mean all axes are moving at maximum speed to get there (even Z).

This is the "odd behavior" that I saw when changing to "Prior-Z" using the controls.

groundcontrol.kv
```diff
@@ -280,9 +280,11 @@
     Button:
         text: 'Open File'
         on_press: root.openFile()
+		disabled: app.data.uploadFlag == 1
```
BarbourSmith: I didn't you could reference data like this...why are we not doing this everywhere?...beautiful!

BarbourSmith:
My computer is installing updates so I am a little behind on merging these, but I will merge as soon as it is finished

groundcontrol.kv
thormj: I'm beginning to like Kivy... but I still get confused about self.x vs root.x vs app.x vs x
I tried root.data, but "root" is the class, and at construct time, it doesn't have "data" (that gets done just a bit later.

And I'm not sure you get to format it.  Text: "0.2f"%root.data.ZPos explodes.

BarbourSmith: I can't tell you how much time I've spent struggling with things not having access to data at construct time. I love kivy, but that was one thing that really bugged me...looks like I just missed the option to call app.data which would have made it easy.

Does it work from within modules too, or just the .kv file?

thormj:
Should I make the "further problem" into an issue?
I tried (no cutting, just running, powered Z) to duplicate it without changing the files on the fly and was unsuccessful... but there was this one time... that it bit me...
But I'm not sure how to get it to bite me again without changing the file on it...
A clunky way to solve it is "when paused gets activated" the Pause button splits into two:
1. Resume Cut (resume the move, but do not feed any more lines) - this should (?) let the red/black crosshairs align (maybe?  I'm not sure what's going on when they differ)...
2. Resume G-Code (resume the sending of g-code)

But... like I said, I don't know what's going on here...  Just when the red and black move apart, bad juju can happen.

BarbourSmith:
I like the idea of making an issue as a place to talk about the issue further.

Gocde can cause unpredictable behavior when you essentially copy and past the second half of one file onto the first half of another file by opening a new file midway through a run.

The real issue is that lines of gcode rely on information sent previously so a line that might not cause any trouble when the machine is in one state can have unwelcome effects when it is in a different state.

I think what I am not understanding is how this PR doesn't fix the issue. If you can't open a new file while one is running, how can the problem happen?

thormj:
I found a way to reproduce it... moving the text for that to a new issue.

groundcontrol.kv
thormj: I've only done it in the .kv file so far... that was slicker than expected....

thormj: On a completely different thread, can you tell me how saving data is supposed to work?  I think Issue #591 is close, but the load/save just isn't.

PR #611, opened 2018-02-06
Some modifications to the Z-Menu

thormj:
So now you can:
  move in absolute terms, 
  jump to the safe height
  jump back to the height you were before the safe height
  jump to zero
  probe for Z-Zero (disabled because there should be a flag/instructions on how to hook that up

Questions:
  1. Should the Z-Zero Probe be in "Measure Machine" (I suppose I was calling it "Setup Machine"
  2. Should the Home button go to Z-Zero?  I know canonically that it should, but if you left it at Z-Safe, there'd be one less prompt for manual Z users (and no touchmarks at 0 for automated z users)?
  3. Is there any interest in an "AirCutting" Flag -- when set, the firmware will ignore all Z changes, but will traverse the paths?  On this dialog, it could be a "Mode: Auto, Manual, Ignored" bit.
  3b. One use I have for it is "oh... it would be cool to add a trench between these 2 points - 
      Pause
      Use G-Code navigator to go to start
      Go
      Pause
      Z-Popup: Go to .25" depth, Set Mode=Ignored, Done
      Resume (cut trench)
      Pause (during cut)
       ...wait for cut....
      Z-Popup: Go to Safe, Set Mode=Auto, Done
      Resume Job

FWIW.... might be an obtuse scenario...

### DIFF
--- a/DataStructures/data.py
+++ b/DataStructures/data.py
@@ -80,6 +80,8 @@ class Data(EventDispatcher):
     zPush = None
     zPushUnits = 'MM'
     zReadoutPos = 0.00
+    zPopupUnits = None
+    zStepSizeVal = 0.1
     
     '''
     Queues
--- a/DataStructures/data.py
+++ b/DataStructures/data.py
@@ -73,7 +73,13 @@ class Data(EventDispatcher):
     iconPath                                              =  StringProperty('./Images/Icons/normal/')
     posIndicatorColor                                     =  ObjectProperty([0,0,0])
     targetInicatorColor                                   =  ObjectProperty([1,0,0])
-    
+
+    '''
+    Misc UI bits that need to be saved between invocations (but not saved)
+    '''
+    zPush = None
+    zPushUnits = 'MM'
+    zReadoutPos = 0.00
     
     '''
     Queues
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -44,7 +44,6 @@ class FrontPage(Screen, MakesmithInitFuncs):
     tick=0
     
     stepsizeval  = 0
-    zStepSizeVal = .1
     
     consoleText  = StringProperty(" ")
     
@@ -314,7 +313,7 @@ class FrontPage(Screen, MakesmithInitFuncs):
     def zAxisPopup(self):
         self.popupContent      = ZAxisPopupContent(done=self.dismissZAxisPopup)
         self.popupContent.data = self.data
-        self.popupContent.initialize(self.zStepSizeVal)
+        self.popupContent.initialize()
         self._popup = Popup(title="Z-Axis", content=self.popupContent,
                             size_hint=(0.5, 0.5))
         self._popup.open()
@@ -325,10 +324,6 @@ class FrontPage(Screen, MakesmithInitFuncs):
         Close The Z-Axis Pop-up
         
         '''
-        try:
-            self.zStepSizeVal = float(self.popupContent.distBtn.text)
-        except:
-            pass
         self._popup.dismiss()
     
     def home(self):
--- a/UIElements/frontPage.py
+++ b/UIElements/frontPage.py
@@ -117,6 +117,9 @@ class FrontPage(Screen, MakesmithInitFuncs):
         self.yReadoutPos    = self.buildReadoutString(yPos)
         self.zReadoutPos    = self.buildReadoutString(zPos)
         
+        #So other widgets can access the position
+        self.data.zReadoutPos = zPos
+        
         #Current Velocity is done here now, not in the firmware.
         self.tick+=1
         if self.tick>=4:    #Can't do this every time... it's too noisy, so we do it every 5rd time (0.1s).
--- a/UIElements/zAxisPopupContent.py
+++ b/UIElements/zAxisPopupContent.py
@@ -146,4 +146,8 @@ class ZAxisPopupContent(GridLayout):
             self.distBtn.text = str(tempfloat)  # Update displayed text using standard numeric format
         except:
             pass                                                             #If what was entered cannot be converted to a number, leave the value the same
+        self._popup.dismiss()
+        
+    def close(self):
+        self.data.units=self.onEntryUnits #Restore original units
         self.done()
--- a/UIElements/zAxisPopupContent.py
+++ b/UIElements/zAxisPopupContent.py
@@ -22,14 +22,13 @@ class ZAxisPopupContent(GridLayout):
         Initialize the z-axis popup
         
         '''
+        self.onEntryUnits= self.data.units
         if self.data.zPopupUnits is None:
             self.data.zPopupUnits = self.data.units
         self.unitsBtn.text = self.data.zPopupUnits
         if self.data.zPush is not None:
             self.zCutLabel = "Re-Plunge to\n"+'%.2f'%(self.data.zPush)
             self.zPopDisable = self.data.zPushUnits <> self.data.zPopupUnits
-        self.setMachineUnits()
-        self.onEntryUnits= self.data.units
 
     def setMachineUnits(self):
         self.data.units = self.data.zPopupUnits #Show the right units on the main screen
--- a/UIElements/zAxisPopupContent.py
+++ b/UIElements/zAxisPopupContent.py
@@ -11,6 +11,8 @@ from   kivy.uix.popup                            import   Popup
 
 class ZAxisPopupContent(GridLayout):
     done   = ObjectProperty(None)
+    zCutLabel = StringProperty("to...")
+    zPopDisable = ObjectProperty(True)
     
     def initialize(self, zStepSizeVal):
         '''
@@ -20,6 +22,9 @@ class ZAxisPopupContent(GridLayout):
         '''
         self.unitsBtn.text = self.data.units
         self.distBtn.text  = str(zStepSizeVal)
+        if self.data.zPush is not None:
+            self.zCutLabel = "to:"+'%.2f'%(self.data.zPush)
+            self.zPopDisable = self.data.zPushUnits <> self.data.units
     
     def setDist(self):
         self.popupContent = TouchNumberInput(done=self.dismiss_popup, data = self.data)
@@ -43,6 +48,16 @@ class ZAxisPopupContent(GridLayout):
             self.distBtn.text = "{0:.2f}".format(float(self.distBtn.text)/25.4)
         
         self.unitsBtn.text = self.data.units
+        
+        if self.data.zPush is not None:
+            self.zPopDisable = self.data.zPushUnits <> self.data.units
+    
+    def goThere(self):
+        '''
+        Go to the position (into work -- you can pop out to a safe height)
+        '''
+        self.data.gcode_queue.put("G00 Z"+ str(-1*float(self.distBtn.text)))
+        
     
     def zIn(self):
         '''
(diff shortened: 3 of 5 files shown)


krkeegan:
Are we really ready to merge this?  This is what the ZMenu looks like.  

image

I love the concept of the functionality, but I am a little skeptical about how intuitive this layout is.  Am I alone on this?

thormj:
That's why I try to stay away from user interfaces when I can.  My UI's are very pick about which users they want to be "friendly" with.  By way of explanation, the "sub buttons" are "subsets" of the bigger buttons (maybe except for the "ToDepth" one, but it "felt" better to keep the "move to absolute" separate from the "relative" buttons.

I suppose another way to handle it would be to split between the "units/zero" column and the "Raise/Lower" column and make an "Absolute" column with "ToSafety", "ToZero", "ToDepth" and "to..." (where ... gets filled in / enabled if you hit ToSafety).

blurfl:
I agree that It might want some rearranging.

blurfl:
How about isolating the 'Absolute Move' buttons from the 'Relative Move' buttons, separated by the big 'Stop' button? A label above each group to indicate the group's function?

thormj:
Ok... see what you think of this one:
image

BarbourSmith:
I like it!

What if we moved the stop button all the way to the right so there wasn't a bump in the top layout? (a nitpick)

thormj:
Ok... how about:
image

krkeegan:
For clarification, what does the To Depth do?

thormj:
Tells the router to go to the depth that is on the left (0.1 inches in this case).
vs To:... which gets filled in with whatever the current depth was when you pressed the "To Safety" button.

thormj:
Ooh... I made it say "To: -0.234":
image

thormj:
Not sure if adding some blackspace helps:
image

krkeegan:
Oh interesting, so does the To button automatically populate with the default value on load?  And the To Depth is always negative I assume?

blurfl:
I wonder whether using the same number input for both relative moves and as a negative value in absolute moves will be confusing? I was thinking of two isolated groups of buttons, but I don;t know how difficult that is. Where does one set the 'To Safety' value?

thormj:
Yep.  To Depth is always negative (since you can't put a negative number in the box) -- and is keeping with the convention that negative numbers are "into the work".  Figuring you'd never need to go to a specific "non-cutting" depth.

The To Safety value is hardcoded in python for now (was borrowed from the "cut calibration pattern"code - 0.25" or 6mm).

Half of me was thinking about coloring the "cutting" buttons rose, and the "safe" buttons (including go to zero) a green shade.  And then duplicating that color scheme in the "manual Z adjustment" popup so you can tell at-a-glance if you should be going to "cut depth" or "no-cut depth".

Then again, adding colors makes it easy to go clowny and my dad is colorblind, so...ymmv.

blurfl:
The 'to...' button - what is it's function?

thormj:
When you "Go To Safety", the current cut depth is saved, and the button enables and indicates that depth.
So it's easy to "pop out" move the sled, and then "pop back in"

blurfl:
"The 'to...' button" might need a different name" - There are two buttons named 'To...' and 'to...'

thormj:
Yep... and the to... will change its label to "To: [saved depth]" when "To Safety" is pressed.

thormj:
Hopefully this is clearer, but it doesn't say "how you save":
image

blurfl:
I was wondering about something along these lines:
'To...' - > 'Plunge to -xx.xx"
'Raise to Safety' -> "Save depth,\r\nprepare to traverse'
'To Zero' -> "Go to Zero"
'to...' _> "Plunge to Previously\r\nSaved Depth: "

blurfl:
I've been looking at the 'Previously Saved' button, it doesn't seem to update when the units change. 

This may be a separate issue, but if I change the units in the z-axis popup while halted during a run, the popup returns with the units changed, affecting the rest of the cut. The popup should save and restore the units that the cut file established.

blurfl:
if I change the units in the z-axis popup while halted during a run, the popup returns with the units changed, affecting the rest of the cut. The popup should save and restore the units that the cut file established.

blurfl:
I opened a separate issue for the z-axis not saving the cut file units.

thormj:
Ok, now this is getting clearer:
image

And:
image

As far as units go, if you change the units from whatever they were when you save it, it disables the "Re-Plunge" button

thormj:
And fixed #615 as well.  If you change units here, that button disables/enables, and you go back to your old units when you leave the dialog.  When you come back, if the units don't match the save point, the button is disabled.

blurfl:
The text size in a button can be adjusted with:
font_size: '12sp'

I believe it's presently using 14sp. We should get this tested on a small screen (Raspi Touchscreen?) When I make the GC window small, The test overflows...

thormj:
I didn't change any sizing in .kv ... but I did split the "Save" button again, and that will help.
I think we should probably go back and decorate all the buttons with sizing information so you could set a system-wide font size.  But it has to be decorated on every button...

blurfl:
That looks pretty good. Unfortunately now when I use the value selector to set raise/lower value, the 'Done' button does not take me back to the Z-Axis popup, the value selector remains on the screen. Moving back to a version from earlier this evening the value selector closes correctly.

thormj:
Sorry... I got my popups and myself mixed up with a2e1002
This should fix everything.

blurfl:
I'm finding some more ways that the present Z-Axis popup behaves in an unexpected manner. It isn't due to any of the changes you propose as users will be using the popup more and in new ways (save-and-traverse/re-plunge) I think I should at least note them.
 Too, since you're the expert on the Z-Axis popup now, maybe you know how to address them 😄.
 I'll open separate issues for them as well as listing them here.

Issue: value can lose its units between popup launches
- Start with the main window in INCHES mode. Open the z-axis popup, enter 0.1 in the value selector. Change to MM, the value correctly changes to 2.54. Leave the popup by means of the 'Done' button. Thanks to the fix you've added to this PR, the main window is still in INCHES mode. Re-open the z-axis popup, the value shown will be 2.54, the units will be inches, the Plunge-to value will be -2.54 (inches). Select MM, the Plunge-to value becomes -64.52. I think the cause is that the numeric value showing in the value selector persists across subsequent uses of the popup, but the units are lost. The fix would seem to be to put the value back into the original units when the popup is exited. 

Issue: popup can exit without using the 'Done' button
- open the z-axis popup, change the units. Note that the main screen units change as well. Close the popup using the 'Done' button, the main screen changes back to the units it had before the z-axis popup opened (thanks 👍💯). If instead you leave the z-axis popup bu clicking outside the window, the units on the main screen do not change back. This leaves the machine interpreting the gcodes in the wrong units.

blurfl:
@BarbourSmith, this is ready for prime time 💯🎉, though @thormj may have some ideas about how to address a couple new issues I've opened.
 Should we merge this and treat those issues separately, or wait to see whether @thormj wants to work his magic on those as part of this major z-axis upgrade?

thormj:
Fixed #616 and #617 and made the zPopupMenu a bit more self-contained.

thormj:
@BarbourSmith so you can put "$.2f"%app.data.zStepSize in the kv file.  But it won't automatically update the underlying data when you write text to it.

And for some reason, if you didn't re-write he data on that button after you changed the units, the units button would be automagically pressed when you got back (so I had to add zAxisPopupConent.py:49 or every time you came back to the dialog the number would multiply)... really not sure why, but I'm glad that fixed it.

thormj:
CRAP.  DON'T TAKE THIS YET.

thormj:
Ok, now we're probably good.  With all the "localized Unit Handling", I never set set the machine to the right units... oops.

Fixed.  And made it so you can't popup the z while sending g-code, because that would be OMG Nooo.

blurfl:
@thormj , sorry to abandon you last night while you were 'on a roll' 😔. The late hour caught up with me.
 I noticed that when I called up the popup from INCHES mode, set .1 as the value, changed the popup to MM the popup value correctly changed and the main units remained INCHES, very good. When I clicked 'Raise', the main units remained INCHES but the Z value began climbing in mm - trying to raise 2.54 inches is a good way to break the sacrificial link in the mechanism. When the movement ended, the Z value changed back to INCHES, and the travel was in face .1 inch.
 I hate to ask for more, but would it be possible for the movement commands to be sent in the units of the main window, even though they are displayed in different units in the popup?
 This may be unnecessary, I'm hoping someone else (@BarbourSmith ?) might chime in with an opinion about this...

thormj:
Is that with 96d180c?  I called a halt to everything when I realized I never sent the G20/G21 commands, and I thought I fixed that in 96d180c... but maybe not.

How do you know what it's going up in?

blurfl:
This is with your most recent commit, yes. Do the steps that I described just above, and watch the value of the Z in the main window behind while you do it.

thormj:
Ah... so in the name of "keeping to myself", I don't mess with the self.data.units when I execute a move from the z-dialog... I manually stuff the G20 / G21 in just before I send the command and reset to the original units after I send the command.  It won't update the other screen  (maybe it should...).

blurfl:
Instead of using G20/G21 to send move values in the units presently selected in the popup, how about sending the move values in the units presently selected in self.data.units, converting them to those units if the popup is presently in the other format?

thormj:
I'd rather send Maslow what was entered so it's very transparent.  I made the zPopup change the self.data.units so it should change the "main screen" during a move.

blurfl:
You're being very patient, and I think I'm not expressing myself well.
 This most recent commit again leaves the main units different from the units defined in the gcode program being cut. That cannot be allowed to happen.

To reproduce:
Open a gcode file written in INCHES
open the Z-axis popup
change to MM in the z-axis popup
close the popup, note the main window is still in INCHES
once again open the z-axis popup, note that the main window has changed to MM
close the z-axis popup, note that the main window is still in MM

 I think the safest way to prevent that is to never change the units of the main window which have been set by the gcode program. Instead, if the popup is to send a movement command, it must change whatever movement value the it is sending into the units set by the gcode program.

thormj:
Sorry, I goofed.  So this one properly saves the units from the main screen, and doesn't change the units unless commanded to move.  If a move is commanded, it changes to the units in the zAxisPopup, sends the G20/21 command, does the move, changes the units back and executes the G20/21 command to go back to the units on entry.

I wanted to do it this way so that whatever is in the "Go To" box is what is sent to Maslow, and if you read the log, you'll see the number you typed into the field.

blurfl:
The movement buttons all send motion commands in the units selected in the popup as you intend. 

When the units change, the 'Plunge to...' button changes its displayed value based on the units selected in the popup, but the 'Re-Plunge to...' button does not. This means that it will re-plunge to the wrong depth if the units displayed differ from those in effect when the 'Save and Raise...' button was used, and that if the units are changed it will again 'Replunge...' to the now different depth matching the displayed value in the new units.

thormj:
Ah... good catch.  I figured since I was keeping track of the units anyway, let's use the units that were saved (instead of the ones on the dialog box).  It /does/ have an odd effect though:
If your machine is in inches, but your z-dialog is in mm, when you Save&GoTraverse, the "Goto cut" will say "Go to xxx IN" (which is where you were, and the units you were in, but /not/ what you have the dialog in).

blurfl:
By 'Goto cut" you mean "Re-Plunge to...", yes? It does show the units it will use when it moves, so it is clear what the button will do.

thormj:
Yep.  for some reason (lack of caffeine?), I used the old terms.
The "Re-Plunge to" will move with the units that were saved when you "Raised to Traverse" and indicates *those units* in its text (no matter what the zPopupDlg or the Machine is in currently), and will go back to "the Current Machine settings" at the end of the Plunge.

BarbourSmith:
Great! I've finally got my computer issues worked out (I think) so I will test and merge today

BarbourSmith:
This looks great to me! Excellent contribution! 👍 👍

PR #613, opened 2018-02-07
Aligns and puts a canvas image onto the background

thormj:
Implements Issue #591 
The settings dialog is beyond crude, but the data is loaded/saved, so it can be customized to individual's machines now.  If it can't auto-pick the points, it will pop up a box and ask you to click the alignment points.


thormj:
Only thing (that I know of) missing is getting main.py:601 to actually work.
I sort-of-faked it by kicking off a timer when you bring up the BackgroundMenu the first time, but this should run periodically from the get-go...

BarbourSmith:
How is this guy coming along? Do you want testing or is it still a work in progress?

thormj:
It's re-ready for consideration but I need help:  The "BackgroundFile" is supposed to be saved, but for some reason it isn't.  Can someone figure out what I did wrong with that part vs the position & HSV tuples (which do get saved)?

BarbourSmith:
The fact that you rebased at all is amazing! Great work!

I will take a look and probably merge it right after the release today. I feel like making this big of a change right before the release is playing with fire, but right after gives us a whole two weeks to dial everything in (and it gives everyone something exciting an new to look forward to!). Great work!

thormj:
I'll agree with slipping it in after this release...  there's a lot of questions I expect there will be in that (like "what are HSV limits", "how do you measure the reflectors", "what kind of reflectors/lights"...

I'll answer things but I won't be able to try anything for 2 weeks (outta town).
Right now - measure with a tape measure and some futzing.
Bigger markers are better, and lights are better than markers (but LED's are right out unless they're in a housing because they're small).
And the HSV limits are in OpenCV format (and that's a little odd in itself); if you replace the TLHSV with "None"  or anything that's not a list, it will always go into "manually pick" mode...

And I'm sure there are going to be more...  I'll try to make up a bunch of "help" pictures, but other than tossing them in the "documentation" folder, I'm not sure what to do with them.

thormj:
Dang.  Rebasing is a total PITA.  But I wanted to use the gCodeRedraw instead of reloading the gcode from disk.

thormj:
Just updating the g-code caused it to move all over the place.  So re-parse the in-memory original file if we need to redraw it.

BarbourSmith:
Is there absolutely no way to do this without open CV? I went to test the idea today and I'm having a little trouble getting open CV installed for python so I feel like maybe we don't want to force everyone on the project to install it. Is it possible to use Kivy's built in image manipulation possibly without the reflective markers which could be difficult to make work for everyone using different cameras in different light levels? I feel like clicking the corners which you suggested might be a better solution.

thormj:
Unfortunately not.  The "stretching" of the picture is done via getPerspectiveTransform and warpPerspective.

what trouble are you having with installing opencv?

thormj:
I installed with the contribs (patented stuff for edge detection, but not used by this code) via
`python -m pip install opencv-contrib-python` , but this should be able to work with the  "more vanilla" `python -m pip install opencv-python`.

BarbourSmith:
Could we do it with just a scatter object like this:

https://youtu.be/4Kctkp7vLAA?t=1m36s

I think adding openCV is really going to be a big headache going forwards with trying to maintain compatibility across all different platforms. I know open CV does the reflector recognition, but I bet most users are going to want to just drag the image to fit on their canvas anyway.

Would it also be possible to break this PR into a number of smaller PRs? I keep trying to find the time to sit down and go through it but it would take a LONG time to fully understand. Changing 20 files and 1,700+ lines of code that we spent months carefully testing and getting to work just seems like too much. It seems like whenever we make big changes like this we end up breaking more things than we fix and it would be much easier if we could review each piece of new code as an individual change rather than as a monolithic chunk.

thormj:
Err... I'm going to close this and reopen it as new branch/pull request.  I just looked at the changes and they're garbage (probably from me not doing a rebase correctly) -- Moved to PR #645.

PR #618, opened 2018-02-09
I added a bunch of notes that I wish I knew when starting

thormj:
Updated the Readme.md file with notes that should help newcomers have an easier time getting started.
In particular, I always get bitten by CamelCase vs camelCase and to self or not to self...


BarbourSmith:
Yes!!! This is wonderful. So much good karma 👍 👍

PR #619, opened 2018-02-10
add Advanced setting for PWM frequency

blurfl:
This builds on Firmware PRs 385, 386 and 389 (all required) to allow changing the PWM prescale frequency from within the Advanced settings.

It offers three choices:
490Hz  ($39=3)
4,100Hz  ($39=2)
31,000Hz  ($39=1)

Note, the $39 values directly correspond to the values used in the firmware System.cpp:setPWMPrescalers()

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -327,6 +327,14 @@ settings = {
                 "desc": "The derivative constant for the velocity PID controller",
                 "key": "KdV",
                 "default": .28
+            },
+            {
+                "type": "options",
+                "title": "PWM frequency for motor control",
+                "desc": "The PWM frequence used for motor speed control",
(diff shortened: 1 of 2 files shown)


krkeegan:
awesome, spectacular use of the settings.  Did all of that make sense when you were working with it?

blurfl:
It was very easy to use, thank you for setting it up!

The section of Report.cpp where REPORT_GUI_MODE is false confused me, each line contains the text of one setting and the value of the next. I see you did it that way to condense things into fewer print statements, easier to see all on one screen once I tumbled to the trick.

BarbourSmith:
I can't wait to play around with these! Awesome work. @blurfl Thanks for pushing the bounds on the PWM/pins used front. When we got things working it would have been easy to say "good enough", but the work you are doing is really taking us beyond that 👍 👍 🙌

PR #620, opened 2018-02-13
Include note requiring restart

reecej:
Include note that changing the setting for the chain wrapping on the top or bottom of the sprockets will require the Maslow board to be restarted.

I'm still working through some of the recent changes, so I haven't been able to get a more descriptive pop-up screen done. But this at least notifies the user that a restart is required.

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -183,7 +183,7 @@ settings = {
             },
             {
                 "type": "options",
-                "title": "Side of Motor Sprockets That Chains Go to Sled",
+                "title": "Side of Motor Sprockets That Chains Go to Sled (Requires Maslow Board Restart)",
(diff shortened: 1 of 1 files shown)


krkeegan:
I think we can fix this from the firmware side.

If setupAxes then kinematics.init and then kinematics.forward is called whenever this setting is changed on the firmware, I think this will solve the issue.

reecej:
That would be great! I'm in the February order group, so I don't yet have a machine to test on. I was going to come back and try to get it working without a restart, but if that would do it then that would be much better! Have you tested that by chance?

BarbourSmith:
Great! I vote +1 for doing it on the firmware side, but I am going to merge this for now because 1.06 comes out tomorrow and we should not that a restart is required until we take care of it on the firmware side

PR #622, opened 2018-02-13
The other zPopup changes broke the calibration.  Fixed.

thormj:
I also made it go to traverse height when "set zero" is pressed to fix issue #595


blurfl:
I wonder whether there should be a mention in the Step8 text that warns the user that the bit will be raised as a convenience by the 'Define Zero' button.
 As far as fixing the issue in Step8, it does what it says on the can 😄👍

BarbourSmith:
Great! 

I agree that a warning could be good if someone doesn't expect it.

PR #626, opened 2018-02-13
Add a try...except catch to velocity calculation

BarbourSmith:
Sometimes this line is failing with a division by zero error...probably because time() is the same as self.lasttime

This could be because the value of last time didn't update properly somewhere or because the function was called back to back and something went wrong on the OS side.

The proposed solution here is to just ignore it when that happens and don't update the velocity for that command.

It's not super elegant.


BarbourSmith:
Fixes #626

krkeegan:
Honestly, I think this is the "python preferred way" at least as I understand it.  The only addition might be only catching the divide by zero error and not all errors, but that is being rather nit picky.

Issue #392, opened 2017-08-18
Change color of gcode visual line after cut


jbnimble:
The visualization of the gcode is always white, it would be neat if the color could change after the line has been executed.

blurfl:
How would lines that will receive multiple passes work?

jbnimble:
Good point, not sure. The idea was to give a quick visual way to tell how much is completed and how much is left to go.

blurfl:
There’s a % complete and line number number on the right. I think the percent is based on line number rather than the time it takes to cut any one line.

jbnimble:
It is certainly a visual eye candy feature, and not required for operation. If the consensus is there are better  things to spend time then I'm fine with closing it unresolved.

BarbourSmith:
I'm going to close this for now because I am worried that it would require a lot of ram when loading large files

PR #635, opened 2018-02-16
rename 'zAxisAuto' to 'spindleAutomate'

blurfl:
PR #458 added a setting to enable/disable spindle control. The setting was named 'zAxisAuto' which seems confusing as it is being used as an on/off control for spindle power. Change the name to 'spindleAutomation' to better reflect its use. Note that the change in GC will match an accompanying PR in Firmware.

This accompanies Firmware PR #392

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -215,7 +215,7 @@ settings = {
             {
                 "type": "bool",
                 "title": "Spindle Automation",
-                "desc": "Should the spindle start and stop automatically based on gcode? Leave off for default stepper control.",
+                "desc": "How should the spindle start and stop automatically based on gcode? Leave off for default external servo control, on for external relay control.",
                 "key": "spindleAutomate",
                 "default": 0,
                 "firmwareKey": 17
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -216,7 +216,7 @@ settings = {
                 "type": "bool",
                 "title": "Spindle Automation",
                 "desc": "Should the spindle start and stop automatically based on gcode? Leave off for default stepper control.",
-                "key": "zAxisAuto",
+                "key": "spindleAutomate",
                 "default": 0,
                 "firmwareKey": 17
             },


blurfl:
On further thought, this setting chooses between 'on/off' and 'servo-control' modes. Is there a better name than 'spindleAutomate' to convey this?
 Too, the description seems misleading:
"Should the spindle start and stop automatically based on gcode? Leave off for default stepper control."
 Both modes achieve on/off control via AUX1, one providing a 5V signal for 'on' and the other by means of a servo control signal where 0 degrees is off, 90 degrees is neutral and 180 degrees is on.

Settings/maslowSettings.py
```diff
@@ -215,8 +215,8 @@
             {
                 "type": "bool",
                 "title": "Spindle Automation",
-                "desc": "Should the spindle start and stop automatically based on gcode? Leave off for default stepper control.",
-                "key": "zAxisAuto",
+                "desc": "How should the spindle start and stop automatically based on gcode? Leave off for default external servo control, on for external relay control.",
```
BarbourSmith: I still feel like I wouldn't understand...or maybe I am just thinking the setting does something different.

The options are `off -> The pin toggles digitally high/low to control a relay` or `on -> The pin is a PWM to set speed`...is that right?

blurfl: That's backwards... if the setting is off (the default) AUX1 has a servo signal that is controlled by Mx codes. If the setting is on, the output is LOW or HIGH as would be used to drive an external relay. Maybe 'auxServoRelay'?
 Servo control is different from PWM. Servos, as used here, are motor/gearchain/crank devices that are driven to specific positions of the crank. While the signal is a controlled pulse width, the width of the pulse represents an angle instead of a speed. A 1.5mS pulse will put the servo at its 'neutral' angle, the middle of its range. Varying the pulse width up or down by 1mS will turn the output by 90 degrees in either direction. Removing the signal leaves the output at its present location for some of the simpler examples. That is the way this servo signal is configured, a brief drive to one of the active positions (0 or 180 degrees) followed by a brief 'return to center' pulse train.

BarbourSmith: Woah, thanks for the explanation...that is way different than what I would have expected. 

I like that we have the servo option, but I'm not sure it should be the default...I vote for just regular on off to be the default...but that's not about this PR 😀

BarbourSmith:
Now that I understand what we're doing 🙄 I think this looks great! I'm glad that text is clarified.

davidelang:
We really need a tristate, not a bool here anyway

on/off and servo are actually not what I would expect to be the most common. I 
would expect a raw PWM output to be the most common.

so change this from being a boolean and start enumerating the different values 
and what they do (who knows, some people may need this to be a digial 
communications mode to talk to a speed controller

blurfl:
Unfortunately PWM isn't available on any PowerControl board AUX pins. These two options satisfy all the currently available output control options.

davidelang:
servo control is PWM, so if it can control a servo position, it can control a 
motor speed via PWM

blurfl:
Yes, anything’s possible with software 😉. In practice, though, the Servo library is doing bit-bang PWM at 50Hz. Would that be suitable for laser power control?

davidelang:
On Sat, 17 Feb 2018, Scott Smith wrote:

> Yes, anything’s possible with software 😉. In practice, though, the Servo library is doing bit-bang PWM at 50Hz. Would that be suitable for laser power control?

itshould be good enough for cutting (lowish speeds, consistant power), but 
probably not great for trying to etch (high speeds, rapidlyvarying power)

but even if you aren't controlling a laser, people will want to control the 
speed of their router

blurfl:
The Servo library pretty much limits the PWM to 50Hz ranging from 2.7% to 12% duty cycle, the standard servo signal range. So long as the PWM control for laser power or router speed can live within those limits...

BarbourSmith:
I think the 2.7% to 12% duty cycle is going to be an issue for a laser, but there is no reason we couldn't add an option using the digitalWrite() function which would give us 0%-100% duty cycle options

blurfl:
I don't see how that would work, unless someone also wrote a bit-bang PWM routine to run on a timer. We're already running into situations where PID calculations don't finish in time before the next timer call, I would hesitate to add more time critical overhead like this.
The TimerOne library only does PWM for the pins controlled by timer1 (pins 11, 12, 13), more's the pity. There _are_ three PWM pins (44,45, 46) on the 36-pin connector of the Mega, though, for board REV1.x...

BarbourSmith:
You are right, I had it in my head that analogWrite() would handle the bit-banging behind the scenes (which would probably still have been too slow), but from the docs:

> On most Arduino boards (those with the ATmega168 or ATmega328P), this function works on pins 3, 5, 6, 9, 10, and 11. On the Arduino Mega, it works on pins 2 - 13 and 44 - 46. Older Arduino boards with an ATmega8 only support analogWrite() on pins 9, 10, and 11.

PR #636, opened 2018-02-17
Keep the raw file in memory and be able to refresh the graphics without reloading the code

thormj:
Fixes issue 633 (not that it shows up badly yet...) and lets you see both the raw and condensed g-code in the view-gcode dialog


thormj:
Err... Moved into PR #645.

PR #645, opened 2018-02-20
Combined/revised patch for Load Image on Background

thormj:
Replaces PR #613 and PR #636, and doesn't pollute self.data as much as the old one.
You must use opencv; install with `python -m pip install opencv-python` as the scatterobject won't stretch things, and the texture_wrap won't stretch things caddywompus.

You might be able to do it with a kivy Transformation matrices, but I don't know how you'd calculate them (opencv can do it with 4 points from source to 4 points on dest).  -- The code that does this is `backgroundmenu.py:229` (warp_image)

**There is a bug:** When you reload_background, it blanks it instead - `backgroundMenu.py:126`
I dealt with this before by putting all the variables into the self.data space, but that seems polluting...
Does anyone know why `self.backgroundFile` shows up empty here when it's still detected by "Tick"?

**And another bug:** The backgroundFile is /not/ being saved/loaded properly.  I patterned it after gcodeFile, but evidently I suck at strings?!

Anyone know why I can't get these 2 to cooperate?

### DIFF
--- a/Connection/serialPortThread.py
+++ b/Connection/serialPortThread.py
@@ -151,7 +151,13 @@ class SerialPortThread(MakesmithInitFuncs):
                 #Send the next line of gcode to the machine if we're running a program
                 if self.bufferSpace == self.bufferSize and self.machineIsReadyForData: #> len(self.data.gcode[self.data.gcodeIndex]):
                     if self.data.uploadFlag:
-                        self._write(self.data.gcode[self.data.gcodeIndex])
+                        #Skip blank lines
+                        while self.data.gcode[self.data.gcodeIndex].isspace() and self.data.gcodeIndex < len(self.data.gcode):
+                            self.data.gcodeIndex = self.data.gcodeIndex + 1
+                         
+                        #Make sure we didn't fall off the end!
+                        if self.data.gcodeIndex <= len(self.data.gcode):
+                            self._write(self.data.gcode[self.data.gcodeIndex])
                         
                         #increment gcode index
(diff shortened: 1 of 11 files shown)


BarbourSmith:
After sleeping on it I think I've got a solution to the open CV issue. What if we make it so that there is a setting option to turn it on and off? With the setting off open CV is not imported and the program runs normally, if it is enabled then open CV must be installed and the background image is supported. Python is nice in that it lets us load libraries dynamically like that.

I am a little bit worried about all the changes to how files are formatted and sent to the machine in this PR, is there any way we could put those in a separate PR? They feel like a different change. Because there are SO many program that generate gcode out there and they all have different quirks (no spaces, "," instead of "."...etc it took us MONTHS to get something that seems to render most gcode files right and I'm worried about breaking it.

thormj:
On OpenCV: I'm certainly game.  Not sure how to do this  with GitHub-- do you patch me and I pull your patch into me, and then ?

As far as the g-code handling goes, an alternative would be to pull PR #636 first and test.  It's actually probably best if we do that and make sure it works... then, I'll take your notes and such and redo this patch?
I do like the explanations I put in the commits of PR #636 and they'd be lost if you just snagged this (otoh, I am now convinced that rebasing is teh evil.--- when you update master, I'll stash this, pull check out master, and then un-stash it and choose which pieces go in carefully instead of as part of a rebase).

So, I've un-closed 636 for your consideration.  The big thing that I did there was leave extra "blank lines" in the g-code (so both copies lined up) only to take them out when sending them.

PR #649, opened 2018-02-22
Create new framework for calibration process

BarbourSmith:
When the calibration process was originally written it was a very linear process which always progressed in the same way. It also has simply evolved a lot over time and had become cluttered and difficult to maintain.

This pull request attempts to re-write the way the calibration process framework works.

Each step is now a self contained widget and they can be loaded, removed, or reordered at any time. Changes to one step will not affect other steps.

Each calibration step has an `on_Enter` function and an `on_Exit` functions which are called when the step begins and when it finishes. When a step is done with whichever part of the calibration process it is working on, it calls the `readyToMoveOn()` function which will remove it from the display and load the next step.

The functionality is mostly in place in this pull request, however only one path through the calibration process is supported right now. The machine must use the chains passing over the top of the sprockets and triangular kinematics. I will add the other possible paths in separate pull requests because this one is already so massive, and because I want feedback and discussion on how to do the chains over the bottom option.

Graphically it looks very similar to how it did before:

image

I've run through everything once, but this shouldn't be considered stable yet. I want to test it quite a bit more before considering merging it. Any feedback is more than welcome.


BarbourSmith:
I'm feeling good enough about these to merge them 😀

PR #653, opened 2018-02-24
Remove Unnecessary Calls to PushSettings and Config.Write()

krkeegan:
Both are covered by configSettingChange().  The calls to write()
are not that annoying other than taking unnecessary CPU cycles and IO.

The calls to pushSettings can create issues when sending multiple
settings to the machine at once.


BarbourSmith:
Are we sure we don't need the write() calls?

The config docs seem to indicate that we do:

https://kivy.org/docs/api-kivy.config.html

But maybe that's just for changing settings related to Kivy (frame-rate...etc?)

krkeegan:
The calls to write are unnecessary because we call write in configSettingChange() which is called everytime a setting is altered.  So it was just duplicative.

PR #658, opened 2018-02-27
Remove Requirement for Restart on Change in Chain Direction

krkeegan:
Should be merged with 

https://github.com/MaslowCNC/Firmware/pull/408

No version number change required.

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -190,7 +190,7 @@ settings = {
             },
             {
                 "type": "options",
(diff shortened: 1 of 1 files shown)


Settings/maslowSettings.py
```diff
@@ -190,7 +190,7 @@
             },
             {
                 "type": "options",
-                "title": "Side of Motor Sprockets That Chains Go to Sled (Requires Maslow Board Restart)",
```
BarbourSmith: ❤️

PR #409, opened 2017-09-05
Corrects typo in step 2

designfrontier:
Closes #406

Nothing else to say. Simple typo fix.

PR #664, opened 2018-03-07
Add Max Feedrate to Advanced Menu

krkeegan:
Not strictly necessary, but given that we are providing a warning to users, we probably want this in GC so that we can help them lower the feedrate if they see warnings.

Also, https://github.com/MaslowCNC/Firmware/pull/409 won't cause an update on the user end, unless they wiped their eeprom or otherwise caused a reset of the settings to their default value.  This change to GC will cause the default value to be pushed to the machine.

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -222,6 +222,14 @@ settings = {
             },
             {
                 "type": "string",
+                "title": "Max Feedrate",
+                "desc": "The maximum feedrate in mm/min that machine is capable of sustaining.  Setting this value too high will cause movements to start before the prior movement finishes.",
+                "key": "maxFeedrate",
+                "default": 700,
+                "firmwareKey": 15
+            },
+            {
+                "type": "string",
                 "title": "Home Position X Coordinate",
                 "desc": "The X coordinate of the home position",
                 "key": "homeX",


BarbourSmith:
I'm glad this is a setting now. Good choice.

If we decide to increase the default is there a way to make that change part of an update or will everyone be stuck at 700 unless they change the setting manually?

krkeegan:
Yes, that will happen, once a user loads this, their ini file will be edited to include the default.  

This has come up with a few other settings.  The prior solution I think has been the PID style default/custom toggle.  That gets a bit bulky in the UI.

Another idea is to allow the settings input to itself to permit a default notation.  So for example, all of the settings would be set to the text default.  We would still describe the default value in the settings text.  If the user wanted to change a setting they could enter the numerical value.  Inside GC we would then know if the user wants to just follow the Maslow recommended values or if they have a specific value in mind.

This is a bit complex on the coding end, but not too bad.  And it would allow the user to try things out and return to the default control by changing the value back to default.

krkeegan:
That last commend may not be entirely clear.  Here is a sample of what I was thinking:

image

In this case, the user wants the Max Feedrate to follow the Maslow recommended default, which may change on GC updates, but wants the Home Position X Coordinate to remain at 0.0 even if there is an update to GC.

davidelang:
we have the config version to help us cover this situation, if you have a new 
config version, and the value of the max feedrate is the former default, change 
it to the new default. If it's something else, we can popup a message telling 
the user that the defaults have changed, and they may want to check their custom 
settings.

BarbourSmith:
> So for example, all of the settings would be set to the text "default". We would still describe the default value in the settings text. If the user wanted to change a setting they could enter the numerical value. Inside GC we would then know if the user wants to just follow the Maslow recommended values or if they have a specific value in mind.

This is absolutely genius. It's clear. It's concise. It's a wonderful solution.

blurfl:
I like the approach of using the text string 'default' to make the value track changes in the default values. I think this is what most users will want to do, and it makes that painless. It has the secondary benefit of highlighting settings that have been changed from the default, to help one remember things tried in the past but not restored.

PR #667, opened 2018-03-12
Add Gero's suggestions for calibration routine

BarbourSmith:
@GeroBH pointed out that the sled may not actually end up in the middle of the sheet during the calibration process because the calibration process doesn't have a good guess for the rotation radius at this point. 

This PR fixes that issue by moving the part where you enter the rotation radius guess before the step where it moves to the center. It also adds a button to command the sled to move to the center instead of doing it automatically when the step is over which was more confusing.

I'm  not going to merge this one until after the 1.09 release because as much as it seems safe, We've all done a lot of testing on 1.09 to make it stable and I would rather not change anything unless we absolutely have to.

### DIFF
--- a/CalibrationWidgets/calibrationFrameWidget.py
+++ b/CalibrationWidgets/calibrationFrameWidget.py
@@ -107,23 +107,32 @@ class CalibrationFrameWidget(GridLayout):
             self.listOfCalibrationSteps.append(setTo12)
             
         
-        #add extend chains
-        measureOutChains                                = MeasureOutChains()
-        self.listOfCalibrationSteps.append(measureOutChains)
-        
-        #add set z
-        adjustZCalibrationDepth                         = AdjustZCalibrationDepth()
-        self.listOfCalibrationSteps.append(adjustZCalibrationDepth)
-        
         if App.get_running_app().data.config.get('Advanced Settings', 'kinematicsType') == 'Triangular':
             #add rotation radius guess
             rotationRadiusGuess                         = RotationRadiusGuess()
             self.listOfCalibrationSteps.append(rotationRadiusGuess)
+            
+            #add extend chains
+            measureOutChains                                = MeasureOutChains()
+            self.listOfCalibrationSteps.append(measureOutChains)
+            
+            #add set z
+            adjustZCalibrationDepth                         = AdjustZCalibrationDepth()
+            self.listOfCalibrationSteps.append(adjustZCalibrationDepth)
+            
             #add triangular kinematics
             triangularCalibration                       = TriangularCalibration()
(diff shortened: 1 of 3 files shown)


blurfl:
Maybe it's wrong to talk about 'the center' before the full calibration has completed. Users seem to get stuck worrying that the sled isn't where they expect it to be, when they need to run the next steps before evaluating accuracy. Asking the user to choose 'the center' is flawed because the calibration process hasn't calculated the sled/linkage/slack variables yet.
I suggest looking at the wording, to make it clear what not to expect at this point. If the sled hangs a long way below the expected center, that is probably a clue that the linkage type or linkage length guess need to be revisited. Even when those are close, the sled won't be in 'the center' until after the last of the calibration steps hashed a chance to run.

BarbourSmith:
That is a GREAT point. Maybe the real issue is the wording. You are spot on (pun intended 😉 ) that we're saying that it should be in the center so we're setting everyone up to feel like something is wrong when it isn't perfect.

What is wording which would make that more clear?

davidelang:
If you manually move the sled to the center before starting the triangular 
kinematics calibration cuts, it's possible to calculate everything without 
measuring the distance from the top of the workpiece to the top cut (and 
therefor, no longer need to enter the bit diameter so that we can add have a bit 
width to the measurement.

fewer manual measurements mean more accuracy :-)


David Lang

blurfl:
Determining ‘the center’ is a measurement in itself. Moving the sled to that point is another. If ‘the center’ doesn’t happen to be on the line midway between the motors, there is yet another point of confusion/dissatisfaction. Better to let the machine have first crack at it.

blurfl:
>What is wording which would make that more clear?

I see that both UIElements/calibrateLengthsPopup.py and UIElements/measureMachinePopup.py talk about 'the center':
>"Pressing Next will move the sled to the center of the work area"

Maybe using 'middle' instead of 'center' would be general enough. I wonder if the accompanying picture might help if the sled were just a little above or below the 'middle', to help avoid the appearance of precision at this stage. Was the picture taken during an initial calibration?

davidelang:
On Mon, 12 Mar 2018, Scott Smith wrote:

> Determining ‘the center’ is a measurement in itself. Moving the sled to that 
> point is another. If ‘the center’ doesn’t happen to be on the line midway 
> between the motors, there is yet another point of confusion/dissatisfaction. 
> Better to let the machine have first crack at it.

I would say "the center" is on a line midway between the motors, just as a 
matter of definition.

I also think you should be able to adjust the four bountries to different 
values, so that if your workpiece is not centered on the machine, you can still 
cover the entire workpiece.

PR #671, opened 2018-03-26
Move choose chain over sprocket direction to after measuring distance between motors

BarbourSmith:
As mentioned in this [forum thread](https://forums.maslowcnc.com/t/calibration-issue/2892/23) asking for the chain over the sprocket direction before measuring the distance between the motors is setting folks up to be confused because measuring the distance between the motors may use the chains in a different way.

Should we move asking the kinematics type and the guess for the vertical distance to the motors down too just to keep them together?

### DIFF
--- a/CalibrationWidgets/calibrationFrameWidget.py
+++ b/CalibrationWidgets/calibrationFrameWidget.py
@@ -45,9 +45,6 @@ class CalibrationFrameWidget(GridLayout):
         intro =  Intro()
         self.listOfCalibrationSteps.append(intro)
         
-        chooseChainOverSprocketDirection             = ChooseChainOverSprocketDirection()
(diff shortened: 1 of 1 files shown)


blurfl:
> guess for the vertical distance to the motors

Is that really necessary? Could we use the current value if it is present, or at least show what it is/was? I keep having to go look it up when running tests 😁

BarbourSmith:
If you skip it it will use the current value. I think it's important to have a chance to enter something because the calibration process will dial it in, but we have to be in the ball park

blurfl:
>If you skip it it will use the current value.

If I click 'Next' without entering a number I bump my nose... "Couldn't convert that to a number..."

blurfl:
Moving the top/bottom choice does what is wanted, and removing the middle picture helps focus the eye on the chain placement. 👍💯

PR #674, opened 2018-03-29
Handle special case where an arc is 360 degrees

blurfl:
This addresses issue #672:
 Fixed the previous trig error, arcLen is in radians instead of degrees...


blurfl:
Closing this again - Atom has trodden roughshod over the whitespace...

PR #675, opened 2018-03-29
Correct rendering error wnen the arc to render is a full circle

blurfl:
This addresses issue #672:
Fixed the previous trig error, arcLen is in radians instead of degrees...

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -327,6 +327,8 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
                 direction = 1
             
             arcLen = abs(angle1 - angle2)
+            if abs(angle1 - angle2) == 0:
+                arcLen = 6.28313530718
             
             i = 0
             while abs(i) < arcLen:


blurfl:
This gcode sequence will test three cases of drawing a full circle and one with an 'almost full' circle:
```
G20 G90
G0 X0 Y0
G3 X.0 Y0.001 I1.0 J0.0 F100  # 1 inch almost-circle
G0 X0 Y0
G3 X0.0 Y0.0 I2.0 J0.0 F100  # 2 inch circle
G0 X0 Y0
G03 I3.0 J0.0 F100  #3 inch circle
G0 X0 Y0
G03 I4.0 F100  # 4 inch circle
M30
```

davidelang:
if you are creating a test, please add a test where the start and end are an 
almost-circle, but only after 4-5 decimal places

David Lang

blurfl:
Driver headspace error.

blurfl:
The PR has been updated to eliminate e-notation and handle parameters with an arbitrary number if decimal digits.
Note that the PR in its present form truncates the display of the parameters at six digits past the decimal point, apparently python 2.7's default. That can be addressed in the format statement introduced in the PR.

PR #677, opened 2018-03-31
Address issue #672   g02 and g03 rendering error

blurfl:
I've added code that handles X and Y parameters with 12 or fewer decimal places, converting the e-notation to a decimal string with the appropriate number of digits for that one parameter. Longer parameters are truncated at 12 places. The fix is centered in the code where changes to the home coordinates combine with the X and Y parameters. The 12 decimal place limit includes those added by the new home coordinates.
 This should correctly handle input from programs that use insanely long parameters, but it doesn't address the issue of string length for a gcode line other than only giving each parameter the number of decimal places it had in the source file. Handling unreasonable input files would be better handled elsewhere.

### DIFF
--- a/UIElements/gcodeCanvas.py
+++ b/UIElements/gcodeCanvas.py
@@ -327,6 +327,8 @@ class GcodeCanvas(FloatLayout, MakesmithInitFuncs):
                 direction = 1
             
             arcLen = abs(angle1 - angle2)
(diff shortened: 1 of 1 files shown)


BarbourSmith:
Woah, this is pretty slick! Great work @blurfl ... way to identify the issue, find an example that replicates it, then solve it 👍 👍

PR #678, opened 2018-03-31
Auto spindle control for triangular calibration

reecej:
This enables automatic spindle control for the triangular calibration process. It just adds the M3 and M5 commands to enable the spindle to be powered on and off during the triangular calibration test cuts.


BarbourSmith:
Looks good! 👍 👍

PR #679, opened 2018-03-31
Incorporate motor spacing into triangular kinematics calibration

reecej:
This adds an additional pair of cuts to the triangular calibration test cut pattern, and uses these cuts to fine tune the motor spacing parameter.

I'm currently putting this through testing on my machine. However, I would appreciate any and all testing, as my machine is just being set up and does not have the usage to confirm if this works as desired.


reecej:
A quick test on this is giving me bad results. I'm checking through the code now, but would appreciate a second set of eyes to see if it's my setup or the code!

reecej:
Just made an update, which is giving me much better results. With this update the motor spacing is tuned better than before, and does help improve the vertical cut accuracy compared to the previous algorithm on my machine.

reecej:
Looking further, I believe the now cut 7 is no longer needed, and the new cut 5 can be used instead. I'll work on that. I think the motor spacing also needs some further tweaking too.

reecej:
After further testing, I'm putting this on hold for the time being. I can't seem to get this to consistently give better results (or in some cases even equivalent results) to the current calibration algorithm. Part of the issue is that errors induced by motor spacing are very similar to errors induced by the rotation radius, making it difficult to tell them apart.

I'm also working through a new sled, so it's possible that others may see better results, but so far my results have given me the conclusion that this isn't ready yet. I may keep working on it in the future and see if it can be improved.

davidelang:
I can't think of any way of detecting errors in motor spacing, it'a a constant 
that let's us figure everything else out.

What we need to do is figure out what's wrong with the current motor spacing 
measurement. Is it that there is an extra thousanth of an inch or so per link 
lenth in the chain (there is _some_ gap between the pin and the hole at each end 
of the link, otherwise it couldn't move, when we pull it tight, do we close that 
gap??)

madgrizzle:
Perhaps, we can have the machine measure the distance and then have the user measure the distance all while its under tension.  The ratio becomes a new calibration factor that is used to offset the distance.  So if the machine measures 3000 mm but the user measures 3005 mm, then the ratio becomes 1.00166667 and when the firmware is calculating how much chain to spool out, it multiplies the number it calculates from the kinematics formula by the ratio (i.e., 1000mm of chain = 1001.66667 of actual chain).

reecej:
I think an important question to answer is, does the current motor spacing measurement induce enough error that it's worth the complexity of trying to tune later on? Part of the purpose of this PR was to see if the machine could be improved by tuning the motor spacing. So far my experience has been that it can't (at least with the code in this PR), but I put this out here so others could try it out too.

davidelang:
In 3000mm of chain, there are almost 500 pin joints, if there is 0.001" of play 
in each joint, that translates into an error of 0.5" (~12mm)

If we have this sort of error on the motor measurement, we can also have the 
same sort of error on the chain length calculations.

I think that if we are within a mm or so on the motor spacing, it's going to be 
good enough, but the reports that we are getting of being off by >10mm worry me.

reecej:
I agree, even small variations in chain tolerance or small amounts of flex in the pin joints could dramatically alter the reading. In the calibration, an error in the motor spacing would result in the rotation radius being calculated wrong. If the motor spacing was significantly off, the algorithm would not be able to complete (and give the user the error message saying the machine could not be calibrated).

I'm working on testing my machine to compare my automatically-measured motor distance to a manually measured value to see how far off I am.

reecej:
I have been working and have some ideas here. I'm going to close this to keep the GitHub repo clean and move the conversation back to the forums.

PR #685, opened 2018-04-06
Add chain tolerance parameters

reecej:
This adds the chain tolerance parameters, calculates the new distPerRotation values, and sends them to the firmware.

This is the first GC update for Firmware [422](https://github.com/MaslowCNC/Firmware/pull/422).

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -189,6 +189,20 @@ settings = {
                 "default": 6.35
             },
             {
+                "type": "string",
+                "title": "Chain Tolerance, Left Chain",
+                "desc": "The tolerance adjustment for the left chain length, in percent",
+                "key": "leftChainTolerance",
+                "default": 0
+            },
+            {
+                "type": "string",
+                "title": "Chain Tolerance, Right Chain",
+                "desc": "The tolerance adjustment for the right chain length, in percent",
+                "key": "rightChainTolerance",
+                "default": 0
+            },
+            {
                 "type": "options",
                 "title": "Side of Motor Sprockets That Chains Go to Sled",
                 "desc": "On which side of the motor sprockets the chains connect to the sled",
@@ -474,6 +488,16 @@ settings = {
                 "type": "string",
                 "key": "fPWMComputed",
                 "firmwareKey": 39
+            },
+            {
+                "type": "string",
+                "key": "distPerRotLeftChainTolerance",
+                "firmwareKey": 40
+            },
+            {
+                "type": "string",
+                "key": "distPerRotRightChainTolerance",
+                "firmwareKey": 41
             }
         ]
 }
--- a/main.py
+++ b/main.py
@@ -158,6 +158,13 @@ class GroundControlApp(App):
         elif (key == 'gearTeeth' or key == 'chainPitch') and self.config.has_option('Advanced Settings', 'gearTeeth') and self.config.has_option('Advanced Settings', 'chainPitch'):
             distPerRot = float(self.config.get('Advanced Settings', 'gearTeeth')) * float(self.config.get('Advanced Settings', 'chainPitch'))
             self.config.set('Computed Settings', "distPerRot", str(distPerRot))
+
+            if self.config.has_option('Advanced Settings', 'leftChainTolerance'):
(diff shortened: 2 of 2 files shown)


davidelang:
A question on this.

With chain tolerance, the circumference of the sprocket is no longer the same 
length as the chain that is released with one rotation of the sprocket, but the 
radius of the sprocket needs to remain the same so that the angled part of the 
chain starts at the same point in space.

does this patch handle this correctly?

reecej:
Good question. I thought about this for a while, and the best solution I was able to come up with was if the chain pitch is actually a bit longer than expected, then it would ride slightly higher on the sprocket and then would exit the sprocket at the slightly larger radius. So for triangular kinematics all calculations treat the sprocket radius as correlating directly with the chain pitch.

Quadrilateral kinematics are unaffected and maintained the original sprocket radius, but the chain length is still updated.

davidelang:
On Fri, 6 Apr 2018, reecej wrote:

> Good question. I thought about this for a while, and the best solution I was 
> able to come up with was if the chain pitch is actually a bit longer than 
> expected, then it would ride slightly higher on the sprocket and then would 
> exit the sprocket at the slightly larger radius. So for triangular kinematics 
> all calculations treat the sprocket radius as correlating directly with the 
> chain pitch.

I don't think this is the case. I think what we will see is the movement of the 
pin inside the hole, so the chain would be the correct length in the sprocket, 
but slightly longer hanging in the air

although the -0.000/+0.001 manufacturing tolerance may mean that the links tend 
to be slightly long, in addition to any slop in the fit.

David Lang

blurfl:
@reecej , in preparing to test this, I want to make sure I understand the settings. In my post [here](https://forums.maslowcnc.com/t/chain-length-differences/1494) I said I found a chain to be 2mm longer than expected. If the default length is 3360mm, the 2mm represents 0.0595% so I would enter 0.0595 as the value for the chain tolerance for that chain, yes?

reecej:
@davidelang I see what you're saying. Let me review the kinematics and see. In either case I would expect a very small discrepancy between the two cases.

@blurfl that's correct! Don't forget the chain calibration after. I should probably add a note saying that is required.

davidelang:
On Thu, 5 Apr 2018, reecej wrote:

> @davidelang I see what you're saying. Let me review the kinematics and see. In either case I would expect a very small discrepancy between the two cases.

I agree, I expect the difference to be minor, but if we are deliberatly ignoring 
someting because it's so small, we should at least call it out in a comment so 
people understand why we did it later.

reecej:
I agree! Furthermore, we're getting to the point where (hopefully) we have all the big error sources ironed out, and are now chasing the small errors to get closer to our desired accuracy level. So aspects like this definitely need to be accounted for properly.

BarbourSmith:
@reecej I am trying to add a process where these values will be computed automatically. My thinking is that we measure the distance between the motors with a tape (probably outside edge to outside edge because it's easier), then do the pull tight measurement on each chain to get their lengths.

What would be the right way to compute the chain tolerance value from those numbers? Is it just a percentage of error?

reecej:
@BarbourSmith that was what I was thinking, and was what I was planning to implement when life calms down.

The chain tolerance is currently a percentage above nominal, so if the nominal pitch is 6.35 mm and the measured average pitch was 6.36 mm then the chain tolerance value would be 0.1575, for example. This was just because I observed several manufacturers listed chain maximum tolerance in percentage, so I was following suit. But if another mechanism of storing this information makes more sense then let's go for it!

BarbourSmith:
Great! That's what I was thinking but I just wanted to check in. A percentage is nice because it's length independent so if we measure 3000mm to be true the distance between the motors and then the chain measures 2990 we get a chain tolerance value of (10/3000)*100 = .33%

reecej:
I agree, the length independent component is nice. But the unfortunate part is this will just give us an average chain pitch variance. We will still have variable chain pitch on a link-by-link basis, and I don't know if a good way to easily determine that more specifically.

The only time this might be a concern is considering how the chain ages. Over time, I would expect the chain links attached close to the sled to stretch more than the other end of the chain, due to usage. So I would expect chain tolerance calibrations to be required on some recurring basis, but for the accuracy to decline over time even with these recalibrations.

Also, for your chain tolerance calculation, I believe it would actually be (10/2990)*100.

PR #689, opened 2018-04-07
Add calibration options

BarbourSmith:
This pull request adds two new calibration options and cleans up the calibration back end a little bit.

### Cleanup:

There is now a standard 'frame' which can hold all the different calibration steps making it very easy to add or shuffle the order of any step in the calibration process

### Adds an option to just run the test pattern

There is now an option under **Advanced** to just run the test cut pattern. This was possible before by skipping through the calibration process, but now it is more accessible 

image

### Adds a manual calibration option

There is now a "Manual Calibration" option which lets the user enter their machine dimensions manually. It is not intended to be a replacement for the the automatic calibration procedure, it is a way to give those who are frustrated with calibrating a quick way to get the machine to a known good state.

image

### Testing needed

This PR needs to be carefully tested because if we're trying to help frustrated people, we need to make sure it works. The things which have changed an need to be tested are:
    

1. The automatic set chain lengths function
1. The manual calibration window
1. The cut test pattern function

Any help with testing those is much appreciated!


blurfl:
I haven't found the manual input screen shown above, but have bumped my nose on various oddities while looking.
Actions/Advanced/RunTriangularTest...' brings up the 'CutTestPattern', from there:
 - 'Back' brings up the 'Congratulations' screen with the 'Finish' button which when clicked takes me back to 'Actions' with 'RunTriangularTest...' showing where 'Advanced' belongs'. Clicking that button reveals the full submenu, most of which work properly except the 'RunTrinagularTest' which has no effect.
 - 'Skip' takes me to the measurement review screen, where the 'LooksGood' button takes me to 'Finished' which does nothing when clicked. 
 - Having visited those places and 'Quit' back to 'Actions' , whether I choose 'Close' and come back to 'Actions' or just go straight to 'SetChainLengths-Automatic' I wind up at the 'RunTrinagularTest' screen instead of the 'SetChainLengths/ZeroSprockets' screen.

 Where to find the manual input screen?

BarbourSmith:
That is very strange. I'm going to look into that right now.

The manual input is screen is hiding in plain sight (there used to be a blank button there so your eye is probably used to skipping over it)

image

blurfl:
I've been through the 'cut test pattern function' function and it operated as expected. 

 Note to self: the chain tolerances should be set to zero before doing this. They are best changed within the Advanced Settings page, if editing groundcontrol.ini the values of distperrotxxxxchaintolerance must be set to 6.35 as well.

BarbourSmith:
> There does seem to be an issue, though. I am set up as 'chainoversprocket = Top', but this dialog changes that to 'Bottom' when 'Enter values' is clicked.

Great point. I was thinking of this as a quick way to help someone who is frustrated so I was trying to just push the most common setting there, but we should give people a choice of chain direction. 

I think it's OK to force triangular kinematics as long as we say that's what is happening in the text, right? If not we need to have an option to enter the distance between the sled attachment points which will make things complicated.

> to add to the confusion, before quitting GC, go to Advanced settings and see 'Top' still selected though groundcontrol.ini says 'Bottom'

Regrettably this one is a bug in Kivy where the settings shown are not updated in the "settings" window except when Kivy launches. I looked into workarounds, but they were all pretty ugly

blurfl:
>force triangular kinematics as long as we say that's what is happening in the text

I hadn't tested that, missed it. How about a warning popup that it mustn't be used with the quad setup? Especially if the changes made won't be visible when the user subsequently visits the Settings screen. Pity about that Kivy quirk. Would [this](https://stackoverflow.com/a/49584456) approach address that? My python and Kivy understanding is very limited...

BarbourSmith:
I like that idea of a warning. Let's be super tricky (Because nobody reads the text 🙄 ) and create what LOOKS like an option to choose triangular or quadrilateral, but with a message that it hasn't been implemented for quad if you click that setting. That way the popup only shows up if it's needed.

That approach to accessing the settings is slick, but I think it's going to have the same issue. The real problem is that Kivy only generates the settings page once when the program loads, and then doesn't update it 😢

PR #700, opened 2018-04-14
Request to add a theme DarkGreyBlue

GeroBH:
Totally selfish and lazy to tweak GC every week I would like check the possibility to add a(my) theme.
screenshot at 2018-04-11 22-30-45
I have hope that it might find fans, but understand if it is rejected as a single persons request.
Kind regards, Gero


BarbourSmith:
Looks good on my end too!

Great work @GeroBH !!! 👍 👍

PR #702, opened 2018-04-15
Fix darwin reset view command

blurfl:
Use dp() to properly scale screen for gcodeCanvas::centerCanvas().

Also introduce a GroundControl setting for the zoom factor in centerCanvas() as used in 'Reset View'. Love that re-worked settings interface!


blurfl:
> just call this function when the settings window closes?

I suppose so, though I was hoping a Kivy-master would stand forth. Remember that the scroll thing happens when scrolling through the textConsole lines as well.

BarbourSmith:
Good point! Let's keep that idea as a backup plan only

PR #707, opened 2018-04-17
Make motor stall warning not trigger on pull chain tight

BarbourSmith:
This disables the new feature which will detect motor stalls during the pull chain tight step when we want to stall the motor in a controled way.

On an side note, it was so cool to see the motor stall warning pop up! I'm really excited about this new feature, I think it's going save a lot of chips :+1:

### DIFF
--- a/CalibrationWidgets/measureDistBetweenMotors.py
+++ b/CalibrationWidgets/measureDistBetweenMotors.py
@@ -119,7 +119,11 @@ class MeasureDistBetweenMotors(GridLayout):
         
         self.originalChainOverSproketDir = App.get_running_app().data.config.get('Advanced Settings', 'chainOverSprocket')
         
+        #pretend we are in the "Top" configuration during this step
         App.get_running_app().data.config.set('Advanced Settings', 'chainOverSprocket', 'Top')
+        
+        #set the threshold for warning that the machine is off target to 200mm esentially turning it off. We dont' want this to trigger when pulling the chain tight
+        self.data.gcode_queue.put("$42=200 ")
     
     def on_Exit(self):
         '''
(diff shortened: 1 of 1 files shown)


CalibrationWidgets/measureDistBetweenMotors.py
```diff
@@ -128,7 +132,9 @@ def on_Exit(self):
         
         '''
         
-        print "measure dist on exit ran"
         
+        #Restore original chain over sprocket direction
         App.get_running_app().data.config.set('Advanced Settings', 'chainOverSprocket', self.originalChainOverSproketDir)
+        #restore all settings to the values stored in the current settings file 
+        self.data.gcode_queue.put("$$ ")
```
BarbourSmith: Do we feel good about me using $$ to reset all the settings to the values stored in settings? It seems to work great

blurfl:
Isn't it '$RS$' or something? I was under the impression that $$ just 'pretty-printed' the values presently stored. GC sends a $$ when it first connects to put it all in the log. I just confirmed that having changed a setting $39=2 with a macro, it no longer matches groundcontrol.ini before or after $$. The $$ shows what is in EEPROM

BarbourSmith:
I thought $$ would pull everything and update anything which didn't match between the two. Here's my terminal readout from the calibration process:

```
Sending: $42=200
Sending: G91
Sending: B09 L3035.0
Sending: G90
Sending: B11 S50 T3
measure left ranSending: B10 L

Read motor spacing: 3011.92
choose sprocket direction on enter ran
Sending: $2=3011.92
Sending: G91
Sending: B09 L10
Sending: G90
Sending: $$
Sending: $42=2.0
Sending: $38=1
```

Although now that I'm looking at it more carefully I'm not sure why the on_enter function looks like it is running before the on_exit. It might be because there is some delay between when the request to measure is sent and when the distance is received...this needs more looking into. Thanks for calling my attention to it

BarbourSmith:
I think $$ is both the command to echo the values in firmware and the command to update them. Here's the comment from when the machine connects

```
       ''' 
        Requests the machine to report all settings.  This will implicitly
        cause a sync of the machine settings because if GroundControl sees a
        reported setting which does match its expected value, GC will push the
        correct setting to the machine.
        '''
        if self.data.connectionStatus == 1:
            self.data.gcode_queue.put("$$")
```

blurfl:
You're right, of course. I proved it to myself late last night and hadn't gotten back here to apologize. 😁

PR #712, opened 2018-04-24
Add Spindle automate options

CptanPanic:
This is the corresponding update that was added to the firmware to allow for options for how the spindle automate flag is set.  This change converts the spindleAutomate bool option into a drop down menu with 4 options, None, Servo, Relay_High, and Relay_Low.


BarbourSmith:
Looks great! You've really been getting things done lately 👍 👍

PR #716, opened 2018-04-26
Add support for buffering on Arduino

BarbourSmith:
We have a buffer on board the Arduino which we can use to store lines of gcode ready to be processed. We stopped using it because way back in the dawn of time it was giving us trouble with some computers loosing data over the connection. At the time we were dealing with a lot of issues and it was easier to simply not use the buffer.

Buffering lines of gcode makes a big difference when executing gcode where there are MANY lines over a short distance. It can make a big difference in reducing the cut time for complex parts.

The pull request adds an option to enable buffering of gcode so that we can start to explore it again.

image


blurfl:
If I remember right, we had stack space issues last time around. Need to be watching that.

blurfl:
Oopsie 😄

PR #719, opened 2018-04-27
Add option to be able to adjust length of chain extended during calibration.

### LINKED ISSUE #701: Original Chain length isn't adjusted based on motor spacing
So I created my maslow with a 12ft beam, and the problem I have is when I extend chains as part of calibration, they aren't long enough.  Looking through the code for the firmware and ground control, it seems that $11 specifies the original chain length, which is 1650mm.  This can be adjusted in firmware, but ground control doesn't adjust this based on motor spacing.  Since this is a parameter in Firmware, I am guessing this would need to be adjusted in ground control when motor spacing is set.  What do you think?

CptanPanic:
Added option so those of us with longer chains can change this value easily.  

closes #701

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -218,6 +218,7 @@ settings = {
                 "default": 1650,
                 "firmwareKey": 11
             },
+            {
                 "type": "string",
                 "title": "Chain Length",
                 "desc": "The length in mm of your chains, used to define the kinematics search space",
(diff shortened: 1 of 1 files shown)


CptanPanic:
Yes the extend chain feature on the Firmware uses the $11 parameter for the length, and previously had to use the macro feature to set this manually.  $11 == originalChainLength  in Firmware.  So no firmware change is needed.

Issue #444, opened 2017-11-02
Calibration process improvements 


BarbourSmith:
I just went through the calibration process for the triangular kinematics and had some ideas for how to make it better.

1) Have an option to enter a guess for the radius of rotation. Right now it just assumes you are using @pillageTHENburn's linkage kit so it takes a lot of tries to end up in the right spot

2) The cut marks move steadily up and to the right with each iteration. They should move up, but zigzag to keep the line roughly centered

davidelang:
Please look at my thread on the topic 
https://forums.maslowcnc.com/t/calibration-pattern-for-triangular-kinematics/879/

for triangular kinematics, we should be able to calculate the correct dimensions 
after one cut.

krkeegan:
I agree with point 1 which I think still needs to be fixed

Point 2 seems to have a slider now?  I would say at least the slider shouldn't default to full bottom which it did for me, as that causes my sled to fall onto the floor.  Perhaps even make it so that users can't cut that low down on the frame even?  Since we can't actually see the test cut placements.

Also getting to .5MM at this distance is way hard for me.  I use an inches tape measure, which is probably no better than 1/16 accuracy, I suspect most ppl will as well.  I eventually gave up aroubd 1/32 of error that seemed to be bouncing back and forth and just lied and told it 75 inches.

davidelang:
If we are measuring things <6" or so, the answer is to go to harbor freight and 
spend $10 to buy an electronic caliper

But the other approach is to try to make the baseline as large as possible so 
that the measurement inaccuracy is a smaller percentage of the baseline.

That's why in the triangular kinematics suggestion, I try to use as much of the 
vertical space as possible.

also keep in mind that people are building differenet size machines, some taller 
(to handle 5'x5' plywood) and some smaller. So the calibration should not assume 
standard machine dimensions. This is why in my triangular calibration 
suggestion, I have the user move the sled to the middle, and then go from there.

blurfl:
Much of this has been addressed in PRs - can this be closed?

BarbourSmith:
I agree. The features I mentioned have been replaced by others. I will close.

PR #721, opened 2018-05-03
Automated chain tolerance computation first pass

BarbourSmith:
This pull request puts in place the first part of a system for computing the chain tolerance values automatically.

Right now the values are not computed but all the measurements are taken.

It's a little rough/fragile at the moment but I want to put it out there so we can get some readings from different folks chains and so I can get some feedback

**Edit:** This PR requires the firmware in PR[#44](https://github.com/MaslowCNC/Firmware/pull/444) to work

**Edit2:** You can launch the process by clicking **Actions -> Advanced**

### DIFF
--- a/CalibrationWidgets/calibrationFrameWidget.py
+++ b/CalibrationWidgets/calibrationFrameWidget.py
@@ -28,6 +28,7 @@ from CalibrationWidgets.finishSetChainLengths               import  FinishSetCha
 from CalibrationWidgets.manualCalibration                   import  ManualCalibration
 from CalibrationWidgets.enterDistanceBetweenMotors          import  EnterDistanceBetweenMotors
 from CalibrationWidgets.measureOneChain                     import  MeasureOneChain
+from CalibrationWidgets.computeChainCorrectionFactors       import  ComputeChainCorrectionFactors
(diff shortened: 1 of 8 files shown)


BarbourSmith:
I just came across a bug in this which is beyond the scope of this pull request, but is probably causing other issues elsewhere too.

The issue came up because I clicked the measure button a bunch of times in a row to see how consistent the results were and it came up 10mm shorter every time I clicked the button...which isn't right because I wasn't moving the chain.

I tried to replicate the bug outside of the new chain testing process and I couldn't get the bug to happen when I would manually send the same commands in the same order. 

After much barking up the wrong trees, I figured out that what was happening was that after the measure command is run, the chain is extended by 10mm so it is slack and ready to remove from the sprockets. Then the next slide is immediately loaded and in the slide transition some new calibration values are sent to the machine.

If new calibration values are sent while the machine is moving and has not had time to power down and save it's position the position is recorded incorrectly and that last 10mm movement is forgotten.

This bug needs to be investigated further and probably fixed on the firmware side. Who knows that mayhem it's causing. For now I've added a 5 second delay which will prevent the next slide from loading before the machine is ready as a work around.

PR #726, opened 2018-05-14
Increase max number of lines to 300,000

BarbourSmith:
The 60,000 line limit is causing more problems than it is preventing. The PR ups the limit to 300,000 essentially removing it

See discussion here: https://forums.maslowcnc.com/t/cutting-speed-doesnt-seem-right/3557/28


blurfl:
If your little tablet can run this, most other platforms should as well. I hope someone on a Raspberry Pi tries it...

PR #729, opened 2018-06-01
use 3 digits precision in z-axis popup

blurfl:
This allows making exactly one revolution on a Ridgid 2202 router with 0.125" travel per revolution.


BarbourSmith:
I feel like I've stopped saying how great your contributions are because they are so regular, but that's not reason to not say thank you! Another great contribution!! Good work 👍 👍

PR #730, opened 2018-06-01
update instructions for MeasureDistBetweenMotors

duderdice:
I found the instructions for this step a bit confusing, so I updated the instructions to reference the specific buttons and better clarify the expected behavior.  Its a minor change but I think folks will find it helpful.


blurfl:
Why the change to the Kivy header? I have much to learn about Kivy.

duderdice:
I'm not that familiar with kivy, but my IDE (Visual Studio Code w/ kivy
extension) indicated the original header was wrong.  I checked the kivy
documentation, and it does indicate that correct header uses a colon after
the hash/crunch symbol.  So I added it to my pill request.

-Darryl

BarbourSmith:
Great work @duderdice! Way to find an issue and fix it. I'm voting 👍 

For anyone following along here's what the new text looks like in GC:

image

PR #733, opened 2018-06-10
Add 360 degree button to set sprocket vertical panel

blurfl:
Calibrating a non-stock z motor is simplified by the z-axis control which allows commanding an arbitrary distance to move. Calibrating a non-stock chain motor is possible using the setSprocketVertical panel of the chain calibration tool, but is presently limited to at most a 5 degree movement per button press.
A clean and easy way to improve this is to add a 360-degree button to the present 5-, 1- and 0.1-degree buttons.

screen shot 2018-06-10 at 12 32 38 pm

### DIFF
--- a/CalibrationWidgets/setSprocketsVertical.py
+++ b/CalibrationWidgets/setSprocketsVertical.py
@@ -14,6 +14,42 @@ class SetSprocketsVertical(GridLayout):
     leftChainLength = 0
     rightchainLength = 0
 
+    def LeftCW360(self):
+        degValue = float(self.data.config.get('Advanced Settings',"gearTeeth"))*float(self.data.config.get('Advanced Settings',"chainPitch"))/1.0;
+        self.data.gcode_queue.put("G91 ")
+        if self.data.config.get('Advanced Settings', 'chainOverSprocket') == 'Top':
+            self.data.gcode_queue.put("B09 L"+str(degValue)+" ")
+        else:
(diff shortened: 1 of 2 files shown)


davidelang:
note that you can also do 'move X distance' for the Z axis

when trying to work out the exact number of steps, you may start out doing one 
full turn, but it's hard to see small errors in one turn, so you will want to go 
to many turns so that any error is magnified.

blurfl:
Testing one, then ten, then 50 and finally 100 rotations using the z axis panel 'Raise' and 'Lower' works well. I found that with higher-speed z motors verifying the number of rotations became an issue - I wound up taking slo-mo video and counting the turns (using tools on hand). This won't be a problem with the chain motors, though, as the lower speed makes watching the turns possible.

davidelang:
On Sun, 10 Jun 2018, Scott Smith wrote:

> Testing one, then ten, then 50 and finally 100 rotations using the z axis 
> panel 'Raise' and 'Lower' works well. I found that with higher-speed z motors 
> verifying the number of rotations became an issue - I wound up taking slo-mo 
> video and counting the turns (using tools on hand). This won't be a problem 
> with the chain motors, though, as the lower speed makes watching the turns 
> possible.

It's less watching the turns than it is checking the starting and ending 
direction.

blurfl:
Both are important. It's a big jump from 10 turns to 50 or 100 and worth checking to see whether that last turn was long or short.

PR #735, opened 2018-07-01
Add .nc files to the macOS package as built by buildozer

blurfl:
buildozer filters the file extensions put into the package, '.nc' was not in the list. As a result, the gcopdeForTesting directory and its files have been omitted...


blurfl:
I need to clean this up before doing the simple one-line PR...

Issue #739, opened 2018-07-19
GC will crash if trying to open "no file"


madgrizzle:
If press select on the file open dialog box without having selected a file first, GC will crash.

In viewMenu.py (def load)

```
  def load(self, instance):
        '''
        
        Load A File (Any Type)
        
        Takes in a file path (from pop-up) and handles the file appropriately for the given file-type.
        
        '''
        
        filename = instance.selection[0]
```
When no file is selected, I suspect instance.selection is not a valid array and trying to access [0] results in "list index out of range".  Not sure the best place to make the fix.  in ViewMenu.py or fileBrowser.py

BarbourSmith:
Great fix! I think we're good to close this issue 🎉 👍 👍

PR #740, opened 2018-07-20
Check that a filename was chosen by FileBrowser

blurfl:
This addresses issue #739 to prevent a crash if the fileBrowser 'Open' button is clicked without a selection. It only provides feedback in the command line, however.
 If no selection was made, the FileBrowser remains open.


BarbourSmith:
Works like a charm! Awesome!

PR #743, opened 2018-07-22
Add text to Connection Timed Out alert

blurfl:
Address two common issues with the serial connection:
•wrong port selected
•firmware not loaded


BarbourSmith:
Brilliant @blurfl ❤️ 👍 👍  Such a simple change that will save people hours of confusion

blurfl:
Improvements or additions to the wording welcome. Are there other causes of the connect/disconnect circle?

BarbourSmith:
Those are the two causes I've seen, but let's keep an eye out and if another cause shows up we can add it to the message right away

Issue #308, opened 2017-06-02
Import plywood image as cutting background


iceblu3710:
I hear rumors of a feature in development so you can save "sheets". I think it would be really handy to be able to import a picture of a piece of plywood on the frame and apply it to GroundControl. That way you could toss any odd shape on and get maximum usage.

PR #751, opened 2018-08-11
[FEATURE] Load an image onto the canvas for easier job positioning

iceblu3710:
While browsing issues I found my original request for this feature #308 as well as thormj's work on the feature #591.

After porting thormj's work to GC1.19 I went out and got some bike reflectors and the auto-align setup was VERY complex and would be a nightmare trying to support with the user base. I decided to remove all auto-align features and drill down to a stable manual-align feature set.

This has been tested extensively by me and I find it IMMENSELY useful. I can toss any old half sheet of stock on my machine and immediately lay out the project without having to tape measure from the corner off the sled with offsets like I was doing.


blurfl:
It seems this won't run without the additional installation of (open)cv2. It brings changes 10 files with 256 additions and 532 deletions, a lot to review. There don't seem to be any instructions for setting up and using it.

 Because of the automated nature of the CommunityGarden, this PR could be auto-committed and break GroundControl. I suggest that the PR be withdrawn or closed until it can be more widely tested and setup and use better documented.

iceblu3710:
To be fair if you select changes from "all commits" its only 22 deletions and only due to me PEP8'ing the functions I had to add pieces into. If I redo those parts I can get it to 0 deletions. I forgot about openCV, I will definitely rewrite the transform and remove the requirement then the only new requirement will be ast module which is part of python.

Does this project have a 2 or 3 phase structure or Experimental, Testing, Stable/Release? If there is a testing branch I can issue the PR to that would be better I agree.

For documentation, how would you like it done? Its surprisingly intuitive and got family members to try it out without explanation and they got it first try. I would add a page to the wiki if the PR was approved.

blurfl:
The project doesn't have a testing branch, but you can invite folks to try your branch on your fork. That's the approach I took with my Teensy35-36 branch of the firmware. After a number of users have helped iron out the code and the instructions, then testing for a PR to the master will be easier. 

Too, there are a lot of changes - are all part of the feature, or are some cleanup and improvements to the existing code? If those can be separated out and individually submitted, that's another way to 'ease into it'. I did that getting ready to release the TLE5206 boards. With all the housekeeping already done, the new feature just sort of drops in.

iceblu3710:
I will open a request for feature feedback in the issues list. Thanks.

PR #755, opened 2018-08-14
Fix G38.2 metric default value

blurfl:
G38.2 defaults to 0.2" movement in inches mode, change the metric default to match. It was mistakenly set for 0.5mm, too small to be of value.


BarbourSmith:
It's awesome to see this process working while I was away ❤️  ❤️ 👍 👍 💯

Issue #752, opened 2018-08-12
[FEEDBACK] Loading a background image onto the work area


iceblu3710:
I have ported thormj's original background work and decided the automatic alignment feature was to complex to make user friendly right off the back. I decided to trim it down to bare bones and put in some error checking and handling.

https://github.com/iceblu3710/GroundControl/tree/BGImage_Kivy

I invite anyone interested to give it a try and let me know what they like and what they would change. Also if there are any bugs on your system. Below is an example picture from my setup and once loaded the benifit of this feature becomes very apparent.

![nf](https://user-images.githubusercontent.com/187394/43997618-d503091c-9d9d-11e8-8d66-b5dff35396b0.jpg)

---

Issues Found:
- None

ToDo
- Remove OpenCV dependency

BarbourSmith:
I absolutly agree that this would be a wonderful feature to integrate into Ground Control.

How difficult do you think it will be to remove the OpenCV dependency? From a glance at the code it doesn't look too deeply critical to the process

iceblu3710:
I didn't think it would be difficult at first but soon realized the whole 2 OpenCV functions are actually very complex in nature. I am working on a pure Kivy implementation now but it's a complete rewrite of the core as the point/translate system is completely different.

This is also my first time with Kivy so it's slow going. From a usability standpoint do you see any changes I should incorporate as I go?

iceblu3710:
@BarbourSmith I have made progress but Kivy widget alignment things are a nightmare. Do you know how to reliably get and set the size of the current window?

The image picker is loaded with a size_hint of (0.9, 0.9) but when I load an image I need to know the root.width during the widget __init__. Binding to size doesn't work as it calls the binding three times. 100,100 then x_actual, 100 THEN x_actual, y_actual.

iceblu3710:
Have a fully working implementation in pure Kivy!

I boiled it down to bare bones with the absolute minimum changes to GroundControl propper. It is probably best to roll out the base concept first. Even a few features I thought were "handy" were actually not relevant enough to include.

#757 should be an easy merge now.

blurfl:
That's much cleaner! What platforms have you tested this on?

iceblu3710:
I have only tested on two Ubuntu machines. 18 & 16, I don't have access to any others.

I really should get a Win10 license, last Windows version I purchased was XP SP1.

Functionality should be stable but I am not 100% positive how it will react to different screen pixel densities. It works on both my ultrawide and laptop though.

PR #757, opened 2018-08-19
Background image on working canvas

iceblu3710:
Working Kivy only implementation to load an image onto the maslow bed. #752 

- Background button added up on the top menu bar. 
    - Open Background: Opens an image picker dialog and then loads the Pick Alignment screen
    - Re-pick Alignment: Uses the current file from the prefernces 
    - Update Background: Re-loads the current file and correction points in case of screen artifacts
    - Remove Background: Removes the current file and points from the config file and clears the screen, GCode is unaffected.

Instructions:

1) Take a picture of your sheet material on your Maslow
nf
1) Click the `Background` button then choose `Open Background`
2) Select the image of your sheet material
3) The `Background PointPicker` will popup
groundcontrol_005
4) The image ratio does not matter, all skewing is corrected by moving the Cyan circles to the corners of your sheet. The red lines should follow the edges of your sheet material.
groundcontrol_006
5) Press `Resize Texture` to get a preview of what you have selected. If you are off on a corner simpily press `Reset Image` and move that point closer. Once you are happy with the result press `Accept`
6) GroundControl will automatically update with the prospective corrected image of your sheet material overlayed on the virtual workspace. 
groundcontrol_007
7) You can now very easily tell if your gcode will fit on that sheet material. This way you can utilize as much material as possible.

### DIFF
--- a/UIElements/backgroundMenu.py
+++ b/UIElements/backgroundMenu.py
@@ -21,6 +21,8 @@ class BackgroundMenu(GridLayout, MakesmithInitFuncs):
     def updateAlignmentInConfig(self):
         self.data.config.set('Background Settings', 'manualReg',
                              self.data.backgroundManualReg)
+        self.data.config.set('Background Settings', 'backgroundfile',
+                             str(self.data.backgroundFile))
         self.data.config.write()
 
     def openBackground(self):
@@ -121,9 +123,6 @@ class BackgroundMenu(GridLayout, MakesmithInitFuncs):
             filename = instance.path
         # Save the file in the config...
         self.data.backgroundFile = filename
-        self.data.config.set('Background Settings',
-                             'backgroundfile', str(self.data.backgroundFile))
-        self.data.config.write()
         # close the open file popup
         self.dismiss_popup()
         # new image loaded so clear manual alignment centers
--- a/UIElements/backgroundMenu.py
+++ b/UIElements/backgroundMenu.py
@@ -1,0 +1,141 @@
+import os
+from kivy.uix.gridlayout import GridLayout
+from UIElements.fileBrowser import FileBrowser
+from kivy.uix.popup import Popup
+from DataStructures.makesmithInitFuncs import MakesmithInitFuncs
+from UIElements.backgroundPickDlg import BackgroundPickDlg
+from kivy.core.image import Image as CoreImage
+from PIL import Image as PILImage
+from io import BytesIO
+import json
+
+graphicsExtensions = (".jpg", ".png", ".jp2",".webp",".pbm",".ppm",".pgm")
+
+
+class BackgroundMenu(GridLayout, MakesmithInitFuncs):
+
+    def __init__(self, data, **kwargs):
+        super(BackgroundMenu, self).__init__(**kwargs)
+        self.data = data
+
+    def updateAlignmentInConfig(self):
+        self.data.config.set('Background Settings', 'manualReg',
+                             self.data.backgroundManualReg)
+        self.data.config.write()
+
+    def openBackground(self):
+        '''
+        Open The Pop-up To Load A File
+        Creates a new pop-up which can be used to open a file.
+        '''
+        # Starting path is either where the last opened file was
+        # or the users home directory
+        if not os.path.isdir(self.data.backgroundFile):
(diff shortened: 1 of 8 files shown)


BarbourSmith:
This is looking great! I'll give it a test right now

UIElements/backgroundMenu.py
```diff
@@ -0,0 +1,141 @@
+import os
+from kivy.uix.gridlayout import GridLayout
+from UIElements.fileBrowser import FileBrowser
+from kivy.uix.popup import Popup
+from DataStructures.makesmithInitFuncs import MakesmithInitFuncs
+from UIElements.backgroundPickDlg import BackgroundPickDlg
+from kivy.core.image import Image as CoreImage
+from PIL import Image as PILImage
```
BarbourSmith: I'm running into an import error with this guy, I'm going to install it but I just wanted to make sure we need it 100%

BarbourSmith:
It looks like Pillow might be the more up to date version of PIL

https://pypi.org/project/Pillow/2.2.1/

iceblu3710:
PIL is Pillow on my system. It seems when Pillow took over they used the same import to make it backwards comparable.

Docs says we can use this to make it load only Pillow:


    import PIL as pillow
    from PIL import Image as PILImage

iceblu3710:
Good catch on the bug. It occurs when there was a file to load in the config but there were no registration points. It by default took the texture of the entire widget. how it was loaded onto the main window AS WELL as the proper raw texture is a bit of a special error inside Kivy.

I simply needed to move the config write to the same routine as when I update the registration marks. Should be golden now.

I borrowed a Win10 laptop from work but I get the OpenGL 1.1 issue on all 4 versions of Kivy I have tried so I can't test this from a windows perspective, only Linux.

iceblu3710:
Bad robot! Apparently, it didn't like the verbiage I used.

BarbourSmith:
Hahah it's not the smartest. It just wait's 48 hours and then closes or merges the pull request. 

I'm still seeing the bug where if I open a background it looks strange the next time I open the window. I'll dig into it on my end

image

iceblu3710:
weird, I was able to reliably re-produce the issue and after `fc7ac95` it was completely gone.

Might need to clear the background settings data out of the .ini but any successful point pick will write to both and any clear or unsuccessful should write defaults of blank to both.

BarbourSmith:
Hmmmm maybe I did something wrong in reproducing it today. I'll try deleting my .ini file and see what happens. I did try picking new points

BarbourSmith:
Yup, deleting the .ini file made the issue go away so I think we're good to go!

Again, great work @iceblu3710 this is a really stellar contribution that I think a lot of folks will be excited about 👍  👍

blurfl:
Wait! Doesn't 'deleting the .ini file made the issue go away' mean re-running the calibration or hand-editing the .ini file.  Wouldn't it be better to fix the issue so that the .ini file doesn't need to be deleted?

BarbourSmith:
I think I just needed to delete the .ini file because my .ini file was generated with the earlier version of the feature. Now I can change the background multiple times or remove it and put a new on in without issue so nobody except me should ever need to delete their .ini file

iceblu3710:
`fc7ac95` moved the only config updates to a routine where both are updated at the same time. This routine is called whenever a reset or failed file pick is called as well as on an accepted point pick.

I could put in a config sanity check but not sure what specific point to enter it. I didn't verify the values in the config write routine as I figured it unnecessary.

How are erroneous values in the config handled in the current code?

BarbourSmith:
We generally treat the values in the .ini file as true and try to catch erroneous values before they are entered in the config file. 

I felt like I tried all the combinations to get bad values in the config file and I haven't been able to get bad values with the new version

blurfl:
So if the user has never used this feature, the proper section is created in the .ini file, yes? That’s good then.

BarbourSmith:
Yes. 

I think the case where we could get into trouble was if the user opened the feature with an earlier version of this PR to test and then exited out of the process mid way through. It was a small corner case that is now handled correctly

iceblu3710:
Good. combining all the config modifications in one place and only calling them twice in total, (fail or success condition) keeps it clean.

Issue #775, opened 2018-09-28
Can't get background image loaded


whoismezero:
The background never updates to show the image selected.

I can open the background popup, I click "Open Background", I can select an image and adjust the corner nodes.  But after I click "Accept" the background never updates.  I see this in the cmd window:
`ERROR  ] Unable to open directory <>
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\kivy\uix\filechooser.py", line 828, in _generate_file_entries
  File "C:\Python27\lib\site-packages\kivy\uix\filechooser.py", line 846, in _add_files
  File "C:\Python27\lib\site-packages\kivy\uix\filechooser.py", line 168, in listdir
WindowsError: [Error 3] The system cannot find the path specified: u''
[ERROR  ] Unable to open directory <>
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\kivy\uix\filechooser.py", line 828, in _generate_file_entries
  File "C:\Python27\lib\site-packages\kivy\uix\filechooser.py", line 846, in _add_files
  File "C:\Python27\lib\site-packages\kivy\uix\filechooser.py", line 168, in listdir
WindowsError: [Error 3] The system cannot find the path specified: u''`

But if I don't "Remove Background" before I close Ground Control, then the non-resized image that I selected is shown in the background.  Clicking "Remove Background" does not remove the image until after the next time I open Ground Control.

blurfl:
It looks like you might be running the development version on a Windows platform. Could you try the [latest release version](https://github.com/MaslowCNC/GroundControl/releases/latest) of GroundControl.Windows-Portable to see if that behaves differently?

whoismezero:
Sorry, I definitely should have mentioned that this is on Windows 10 and this is the 1.21 version of GroundControl.Windows-Portable.  I just downloaded and tested with the latest 1.22 this evening and the issue persists.
I also tried with the 1.20 version since it was still downloaded on my computer, and it works there!

blurfl:
Let's see if I can summarize:
The 'Accept' button closes the background setup screen but does not cause the image to display. Quitting and re-starting GC at this point displays the image.
Similarly, the "remove Background" closes the background setup screen, but the image is still displayed until GC is quit and re-started. 
That's a bug, for sure!

BarbourSmith:
It looks to me like the issue is that for some reason the file path is being set to ""

[...]

PR #776, opened 2018-10-02
Trigger the bound function to the gcode to run on update even if there is no gcode

BarbourSmith:
This is needed to fix issue #775 which was causing a bug where the background image didn't behave correctly when there was no gcode file open because the file updateGcode function opts out if there is no file, but we need the refresh process to run


blurfl:
What do you think about merging this one before the robot times out, to get it into tomorrow’s release?

PR #781, opened 2018-10-19
Add catch for KeyError exception

### LINKED ISSUE #780: click and drag outside the Z window crasches GC
As posted here https://forums.maslowcnc.com/t/zaxis-gc-close-out/7274
there is a bug crashing GC, confirmed with Windows and Ubuntu.
With GC 1.23 and back to GC 1.03

To replicate:
Open the Z menu, click and drag outside the Z window.

Credit for finding goes to TalleyMakeWork

BarbourSmith:
This exception when a motion was detected while a popup was open was causing the program to crash. No action is needed for this exception so it is now captured silently.

Fixes #780

Please test and up vote this pull request if the fix seams reasonable and works for you!

### DIFF
--- a/UIElements/modernMenu.py
+++ b/UIElements/modernMenu.py
@@ -146,11 +146,14 @@ class MenuSpawner(Widget):
             return super(MenuSpawner, self).on_touch_down(touch, *args)
 
     def on_touch_move(self, touch, *args):
-        if (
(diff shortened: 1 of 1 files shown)


blurfl:
That was quick! Works as described on the box..

GeroBH:
Tested on linux works like a charm.
Master speed solving!

PR #788, opened 2019-01-23
Update the picture of the sled attached

BarbourSmith:
As recommended in this thread https://github.com/MaslowCNC/GroundControl/issues/785#issuecomment-456265337 by @guidobender


BarbourSmith:
It usually shows the changes to the image but it's not doing that this time so here is the difference:
image

PR #789, opened 2019-02-15
Change default extend chain length to 1651 mm

madgrizzle:
1651 mm is an integral of 6.35 mm (the pitch of a #25 chain) and therefore would result in a tooth being at 12 o'clock after the chains are extended.  Using this value can allow someone to manually reset the chains without having to go through the long, tedious process of having the motor feed out the chain.


GeroBH:
168 hours = 7 days, 7 thumbs up
stuck 29 days? Why?

BarbourSmith:
Hmm I'm not sure but I will find out why!

PR #794, opened 2019-06-20
Clean up serialPortThread() connection handshake

blurfl:
• reset Arduino Mega by toggling DTR instead of open/close/open the port
• send ctrl-X to trigger a soft reset for non-Arduino boards

Teensy3.5/3.6 and GrandCentralMetroM4 don't reset when GC does the special Arduino serial-port-reset trick. This adds a one-character command to trigger a soft reset. GRBL and MaslowDue use ctrl-X, so we match that.

There is a matching Firmware PR#513 to handle the ctrl-X.

### DIFF
--- a/Connection/serialPortThread.py
+++ b/Connection/serialPortThread.py
@@ -114,12 +114,18 @@ class SerialPortThread(MakesmithInitFuncs):
             msg = ""
             subReadyFlag = True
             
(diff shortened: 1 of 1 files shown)


blurfl:
@BarbourSmith , this PR seems to be stuck in the works - is the Robot on a holiday? 🤖🏝

BarbourSmith:
I'm not sure what is going on there. The robot is merging other things...I wonder why it got stuck on this one. Thanks for pointing that out.

PR #796, opened 2019-07-21
bump version number

blurfl:
bump...

### DIFF
--- a/DataStructures/data.py
+++ b/DataStructures/data.py
@@ -26,7 +26,7 @@ class Data(EventDispatcher):
     
     #Gcodes contains all of the lines of gcode in the opened file
     gcode      = ObjectProperty([])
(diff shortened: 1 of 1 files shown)


blurfl:
The robot's still not honoring the 48 hour limit set in a previous PR... I'm going to merge this one to keep synced with Firmware.

BarbourSmith:
Yeah, the robot really seems to be on the fritz. I'm not sure what's up with it. I have a long term plan to replace all that infrastructure, but things are pretty slow right now so let's just go back to merging things by hand for now. I will try to keep an eye on them too, but you should always feel free to merge anything you would like.

In the longer term I would like to replace the robot with a github app which would be the right way to do something like that, but first I need to get Create working well enough that people can use it to replace the garden because I think that matters more to most people

Issue #799, opened 2019-08-21
The Advanced/"Chain Sag Correction Value for Triangular Kinematics" default value doesn't show


blurfl:
Unneeded text at the end of the title of this setting in maslowSettings.py writes over the default value in the settings screen. Removing the extra text reveals the default value, which is displayed separately from the title.

BarbourSmith:
Nice catch 👍 👍

PR #800, opened 2019-08-21
address issue #799

blurfl:
Unneeded text at the end of the description of this setting in maslowSettings.py writes over the default value in the settings screen. Removing the extra text reveals the default value, which is displayed separately from the title.


blurfl:
Giving the robot a poke, he's snoozed through the 48-hour merge alarm.

PR #801, opened 2019-08-25
add Run/Pause/Stop/FAKE_SERVO hotkeys

blurfl:
- The Run/Pause/Stop hotkeys call the same routine that clicking the corresponding button calls. That handles updating the text on the ‘Pause’ button when it changes state.
 - The key assignments are editable in Settings/GroundControlSettings, though that limits us to one key per action. Defaulted to 'r', 'p', 's' (lower case)
 - The escape key can be assigned using the word escape, the spacebar using the word spacebar. Avoid upper case, shifted or meta characters.
 - Assign a character to disable FAKE_SERVO mode, press that key along with 'ctrl' or 'alt' or 'cmd' to enable FAKE_SERVO. Defaulted to 'f' for off.

### DIFF
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -314,7 +314,7 @@ settings = {
             {
                 "type": "string",
                 "title": "Chain Sag Correction Value for Triangular Kinematics",
-                "desc": "The scaled value computed by the calibration process to calculate chain sag based on sled weight, chain weight, and workspace angle\\ndefault setting: %s",
+                "desc": "The scaled value computed by the calibration process to calculate chain sag based on sled weight, chain weight, and workspace angle",
                 "key": "chainSagCorrection",
                 "default": 0,
                 "firmwareKey": 37
--- a/Settings/maslowSettings.py
+++ b/Settings/maslowSettings.py
@@ -484,6 +484,34 @@ settings = {
                 "desc": "Zoom scale for 'Reset View' command.",
                 "key": "viewScale",
                 "default": ".45"
+            },
+            {
+                "type": "string",
+                "title": "Run",
+                "desc": "Pressing this key is the same as clicking the 'Start' button. Note combinations of keys like \'shift\' + \'=\' may not work as expected. Program must be restarted to take effect.",
+                "key": "runKey",
+                "default": "r"
+            },
+            {
+                "type": "string",
+                "title": "Pause",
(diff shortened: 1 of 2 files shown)


BarbourSmith:
Nice work! It seems like the robot is snoozing on this repo entirely so we'll probably need to merge this one manually as well

blurfl:
I don't know json formatting, but I wonder whether 
> "VoteHours": "48"

should be 

> "VoteHours": 48

Instead?

jbarchuk:
Maybe it's accepting quotes, but failing to parse properly. If it's a
generic error others would have to have noticed it already.

Tried jsonlint.com? Maybe another error, but the 48 shows up as the symptom.


https://stackoverflow.com/questions/15368231/can-json-numbers-be-quoted/15368259#15368259

blurfl:
Thanks for the links. This repo is different from the Firmware repo, having the "VoteHours" key in ROBOT.md  
 The [robot.py](https://github.com/MaslowCommunityGarden/Community-Garden-Robot/blob/master/Scripts/robot.py) script in the [Community-Garden-Robot](https://github.com/MaslowCommunityGarden/Community-Garden-Robot) usesthat to decide when to close voting. This is a recent change, and wants some massaging. 
 Your helpful comment suggests that you grok reading json with python, does anything stand out to you in that robot.py script?

jbarchuk:
I know literally 0.1% about json, py, js. I know enough that I get
different effects with/without digit quotes. I give up and quote everything
because it's easier than thinking. And it's supposed to work.

"...having the "VoteHours" key in ROBOT.md The robot.py
<https://github.com/MaslowCommunityGarden/Community-Garden-Robot/blob/master/Scripts/robot.py>
 script in the Community-Garden-Robot
<https://github.com/MaslowCommunityGarden/Community-Garden-Robot> usesthat
to decide..." Meaning  MaslowCommunityGarden
<https://github.com/MaslowCommunityGarden>/ROBOT.md has the hours? Yes I
see robotURL in robot.py. But in ROBOT.md I see only one line and it's not
about hours.

BarbourSmith:
One should be read as a number, while the other is read as a string. I think we want the number, so that is a mistake in quotes....I think. But because we are using the default value of 48 I think we can safely take that line out entirely. Nice spot. I am going to open a pull request to remove that line...which we will need to merge manually 😁🙄

BarbourSmith:
Wooooo 🎉🎉🎉

PR #805, opened 2019-10-08
Updating

tekweaver:
updated the print function.

regularized tabs and spaces

### DIFF
--- a/CalibrationWidgets/chooseChainOverSprocketDirection.py
+++ b/CalibrationWidgets/chooseChainOverSprocketDirection.py
@@ -17,7 +17,7 @@ class ChooseChainOverSprocketDirection(GridLayout):
         This function runs when the step is entered
         
         '''
-        print "choose sprocket direction on enter ran"
+        print("choose sprocket direction on enter ran")
     
     def setChainToTop(self):
         '''
--- a/CalibrationWidgets/measureDistBetweenMotors.py
+++ b/CalibrationWidgets/measureDistBetweenMotors.py
@@ -93,7 +93,7 @@ class MeasureDistBetweenMotors(GridLayout):
     def readMotorSpacing(self, dist):
         dist = dist - 2*6.35                                #subtract off the extra two links
 
-        print "Read motor spacing: " + str(dist)
+        print("Read motor spacing: " + str(dist))
         self.data.config.set('Maslow Settings', 'motorSpacingX', str(dist))
         
         #put some slack in the chain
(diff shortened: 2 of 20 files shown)


BarbourSmith:
Updating these print functions is a part of a larger switch from python 2.x to python 3.x which I think could have big impacts requiring everyone to update their copies of python. I'm not saying it's a bad idea, but I think it is something we should be conscious about. 

Does everyone feel good about moving in this direction towards python 3.x? Any objections?

blurfl:
How about keeping the python3 changes in a separate branch for now? These print functions wouldn't break anyone's current setup, but many further changes will. When the python3 branch is ready for prime time, offer it as well and let users migrate at their own pace.

BarbourSmith:
Great suggestion @blurfl . I will make a new branch for a Python 3.x port which we can start working towards.

BarbourSmith:
@tekweaver Would you be willing to make a pull request to the Python 3.x branch until everyone is ready to make the move to Python 3.x?

https://github.com/MaslowCNC/GroundControl/tree/Python-Version-3.x-

davidelang:
I believe that these changes can be made under 2.7 with the 'import from future' 
feature of python.

I think it would be far better to do the migration in the main branch than in a 
separate branch that will need all other changes applied to it until we switch 
over.

(plus the is the possibility that we end up depriciating GC in favor of WC as 
the kinks get worked out there)

David Lang

tekweaver:
I don't mind pulling from a different branch. I noticed kivy was complaining about python2.7 being depreciated by January 2020, I needed to do some things to get it working on my box, and I wanted to learn some more than basic(dos) programming. Figured python(3) was a good starting point. 

With that said, please bear with me while I work out the git kinks. Like the message I put on the commit -m :)

davidelang:
to be clear, my concern isn't over your work, it's the problem of how your work 
will conflict with any work other people are doing, and how hard it will be to 
merge things together later.

There are a number of things that changed with python 3, but there are 
transition mechanisms available so that code can be written that will run on 
both python 2 and 3.

what I am fearful of is that you will spend months working on a python 3 branch, 
and then when we try to merge in all the things that have changed on the python 
2.7 branch in the meantime, we run into big problems.

I would rather see a series of changes on the python 2.7 branch that move it 
towards python 3 compatibility while still having all other changes hitting it.

for example, one PR would be to just change the print statements, while 
retaining 2.7 compatibility

David Lang

PR #817, opened 2020-06-19
Add RefineCalibrationParameters folder to adjust the input parameters…

domedave:
RefineCalibrationParameters use for adjust the input parameters based on measured output results. Now it can run with general Terminal command (since I executed on ubuntu). hope that it can embedded on some menu button.


BarbourSmith:
This looks fantastic! I cant wait to play with it. Thank you!

PR #819, opened 2020-09-13
Update INSTRUCTIONS.md

Jwolter0:
Added caveat that GroundControl is discontined.


GeroBH:
Although it feels like something inside of me is breaking, the truth is not to be denied.
Rest in peace GC.
The countless hours we spent together will be remembered till the end of time.

BarbourSmith:
It's not sad! It got us so far and has been replaced by better options! 

@Jwolter0 I'm actually @barboursmith on here 😔

